PR #2, opened 2017-02-24
Added PEP

vasili-v:


pep/client.go
```diff
@@ -0,0 +1,87 @@
...
+func NewClient(addr string) *Client {
+	return &Client{addr: addr}
+}
+
+func (c *Client) Connect(timeout time.Duration) error {
+	if c.conn != nil {
+		return ErrorConnected
+	}
+
+	conn, err := grpc.Dial(c.addr, grpc.WithInsecure(), grpc.WithBlock(), grpc.WithTimeout(timeout))
```
johnbelamaric: Fine for now but this will have to change to allow TLS. Also, the WithBalancer option https://github.com/grpc/grpc-go/blob/master/clientconn.go#L130 

Why the WithBlock? Seems unnecessary here.

PR #12, opened 2017-03-28
Fixed dissapearing obligation

vasili-v:
The issue occurs because `func (c *Client) Validate(ctx context.Context, in, out interface{}) error` doesn't take `Response` structure from "github.com/infobloxopen/policy-box/pdp-service" as "out" argument. It takes any structure and process it similar to json.Unmarshal. So I added following response structure:
```go
type Response struct {
	Permit   bool   `pdp:"effect"`
	Redirect net.IP `pdp:"redirect_to"`
}
```
Validate function uses "pdp" hints to translate protobuf response. Here it sets `Permit` to `true` only if response is "Permit" and `Redirect` to content of "redirect_to" obligation making validation that redirect_to is really IP address. If there is no "redirect_to" obligation in response field `Redirect` retains its previous value (in the case it's `nil`).

### DIFF
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -51,6 +51,11 @@ type PolicyMiddleware struct {
 	ErrorFunc func(dns.ResponseWriter, *dns.Msg, int) // failover error handler
 }
 
+type Response struct {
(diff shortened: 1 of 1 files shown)


yuewko:
LGTM!

PR #17, opened 2017-04-11
Added radix network tree and made selector use it

vasili-v:


pdp/attribute.go
```diff
@@ -89,13 +89,13 @@ func (v AttributeValueType) describe() string {
 	case DataTypeSetOfNetworks:
 		items := []string{}
 		i := 0
-		for _, n := range v.Value.([]net.IPNet) {
+		for n := range v.Value.(*SetOfNetworks).Iterate() {
```
johnbelamaric: What happens if v.Value has the wrong type? Shouldn't you be using the x, ok := type assertion idiom here?

pdp/attribute.go
```diff
@@ -203,20 +203,20 @@ func ExtractSetOfStringsValue(v AttributeValueType, desc string) (map[string]boo
...
 
-func ExtractSetOfNetworksValue(v AttributeValueType, desc string) ([]net.IPNet, error) {
+func ExtractSetOfNetworksValue(v AttributeValueType, desc string) (*SetOfNetworks, error) {
 	if v.DataType != DataTypeSetOfNetworks {
 		return nil, fmt.Errorf("Expected %s as %s but got %s",
 			DataTypeNames[DataTypeSetOfNetworks], desc, DataTypeNames[v.DataType])
 	}
 
-	return v.Value.([]net.IPNet), nil
+	return v.Value.(*SetOfNetworks), nil
```
johnbelamaric: ditto

pdp/network_tree.go
```diff
@@ -0,0 +1,146 @@
...
+		net.CIDRMask(0x19, 32),
+		net.CIDRMask(0x1a, 32),
+		net.CIDRMask(0x1b, 32),
+		net.CIDRMask(0x1c, 32),
+		net.CIDRMask(0x1d, 32),
+		net.CIDRMask(0x1e, 32),
+		net.CIDRMask(0x1f, 32),
+		net.CIDRMask(0x20, 32)}
+
+	ErrorIPv6NotImplemented = fmt.Errorf("IPv6 and mixed network sets haven't been implemented yet")
```
johnbelamaric: How much effort to support it? I'd rather not build up debt.

pdp/attribute.go
vasili-v: Any AttributeValueType instance should be created in the way that its DataType and Value fields are consistent otherwise it means issue with some of parsers we provide as part of pdp or pdpserver. The only option we have in the case - panic.

pdp/network_tree.go
vasili-v: The issue here that the bitradix package supports only 32 and 64 bits keys for IPv6 we need 128 bits. It can be implemented using nested 64 trees and requires to pay attention on different corner cases which results from cross tree address masks. I'd say it may take whole day so prefer to do it later. We can add the task to next sprint.

PR #20, opened 2017-04-21
Added validation of middleware response

pankaj-chhabra:
I am not able to compile CoreDNS. This PR is -

1. To confirm that Code and Unit Test approach is done as expected
2. Will make the code changes as per review comments and check with team regarding how to build and test.

Limitations:
1. Current for response.Redirect in Unit test, address field is not checked


contrib/coredns/policy/policy.go
```diff
@@ -20,6 +20,7 @@ import (
 	ot "github.com/opentracing/opentracing-go"
 
 	"golang.org/x/net/context"
+	"github.com/pkg/errors"
```
johnbelamaric: Use fmt.Errorf instead of this.

contrib/coredns/policy/policy.go
```diff
@@ -52,7 +53,7 @@ type PolicyMiddleware struct {
 }
 
 type Response struct {
-	Permit   bool   `pdp:"effect"`
+	Permit   bool   `pdp:"Effect"`
```
johnbelamaric: I think you need a rebase

contrib/coredns/policy/policy.go
```diff
@@ -149,7 +178,29 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 	}
 
 	if response.Permit {
-		return middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
+		status,err := middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
```
johnbelamaric: You will need to implement your own ResponseWriter that captures any response written to "w". It will look something like this `response` struct here: https://github.com/coredns/grpc/blob/master/middleware/grpc/grpc.go#L103

Another example is https://github.com/coredns/coredns/blob/master/middleware/pkg/dnsrecorder/recorder.go, but that doesn't quite do what we want, because it captures the response but ALSO writes it back to the client. 

Instead, you will create a new ResponseWriter (it can take LocalAddr, etc. from "w"), that will work like the one in the grpc code and capture the dns.Msg. So, after this call you will have the dns.Msg that *would have been* sent back to the client - let's call that `upstreamResult`. Then, you will need to find the Answer section in that `dns.Msg` and read the "address" out of it. That will be packaged and sent to the PDP Validate method. 

Based on the result of the Validate, you will modify (or not) `upstreamResult`, and write it out to w via w.WriteMsg().

contrib/coredns/policy/policy.go
```diff
@@ -119,11 +120,39 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
 	return attrs, foundSourceIP
 }
 
+// This function validates the middleware response
+// Inputs is *PolicyMiddle, *Response and error
+// Output is DNS status and error
+func ValidateMiddlewareResponse(p *PolicyMiddleware, w dns.ResponseWriter, r *dns.Msg,response *Response, status int, err error) (int, error) {
```
johnbelamaric: Validate is not the right name here, it's doing more than validating, it's actually modifying the request if needed. I think just put this into ServeDNS, it's only ever called once.

contrib/coredns/policy/policy.go
```diff
@@ -119,11 +120,39 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+
+	if err != nil {
+		log.Printf("[ERROR] Policy validation failed due to error %s\n", err)
+		return dns.RcodeServerFailure, err
+	}
+
+	if response.Permit == true && response.Redirect == nil {
+		log.Printf("[INFO] Permitting request")
+		return status, nil
```
johnbelamaric: Based on the description below, here is where you would have to do w.WriteMsg(upstreamResult).

contrib/coredns/policy/policy.go
```diff
@@ -119,11 +120,39 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+	}
+
+	if response.Redirect != nil {
+		log.Printf("[INFO] Redirecting request")
```
johnbelamaric: Fine for now, but in the real deal, we will only log errors.

contrib/coredns/policy/policy.go
```diff
@@ -119,11 +120,39 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+		return p.redirect(response.Redirect.String(), w, r)
+	}
+
+	if response.Permit == false {
+		log.Printf("[ERROR] Policy response validation failed due to error %s\n", err)
```
johnbelamaric: This isn't an error. Just a refusal. No need for this clause.

contrib/coredns/policy/policy.go
```diff
@@ -149,7 +178,29 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
 
 	if response.Permit {
-		return middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
+		status,err := middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
+		if err != nil {
+			return status, err
+		}
+
+		responseattrs := []*pb.Attribute{&pb.Attribute{Id: "type", Type: "string", Value: "response"}}
+		if len(r.Question) > 0 {
```
johnbelamaric: I don't think we need this whole section, from here down to line 196.

contrib/coredns/policy/policy.go
```diff
@@ -149,7 +178,29 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+			responseattrs = append(attrs, &pb.Attribute{Id: "domain_name", Type: "domain", Value: strings.TrimRight(q.Name, ".")})
+
+		}
+		if foundSourceIP {
+			responseattrs = append(attrs, &pb.Attribute{Id: "proxy_source_ip", Type: "address", Value: state.IP()})
+		} else {
+			responseattrs = append(attrs, &pb.Attribute{Id: "source_ip", Type: "address", Value: state.IP()})
+		}
+		responseattrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: "Dummy"})
+
```
johnbelamaric: You need to look through the upstreamResult Msg and get the "address" out of the A or AAAA Answer section, and then send that as an attribute too.

policy_id will have to come out of "response".

contrib/coredns/policy/policy_test.go
```diff
@@ -0,0 +1,105 @@
...
+	dnsMsg := new(dns.Msg)
+	//
+	// Case 1
+	//
+	var response1 Response
+
+	// Case 1a - Err is set, return should have err and RcodeServerFailure
+	err := errors.New("Invalid error")
+	status := dns.RcodeServerFailure
+	returnedStatus, returnedError := ValidateMiddlewareResponse(&p, dnsResponseWriter, dnsMsg, &response1, status, err)
```
johnbelamaric: We need unit tests for the whole ServeDNS method and other methods. This is probably a stretch for this PR.

I am thinking is that you will make a pep "Client" interface, and change this struct:

https://github.com/infobloxopen/policy-box/blob/master/pep/client.go#L25

to a "client" (lowercase) that implements the Client interface.

Then, in maybe a "test" directory you will add another implementation "testClient", which also has a "SetResponse" method. Then you'll need to set the "pdp" in the policy middleware to this test client. Then, you can make calls to SetResponse to tell the testClient what to reply with just before the test, then call through the ServeDNS method to test all the logic. You'll also need to do similar with the "Next" middleware.

All this can wait though, don't spend a ton of time on it now.

contrib/coredns/policy/policy.go
pankaj-chhabra: Changed code to use fmt.Errorf, removed import of github.com/pkg/errors

contrib/coredns/policy/policy.go
pankaj-chhabra: Will do before the next pull request and resubmit a new PR

contrib/coredns/policy/policy.go
pankaj-chhabra: Had moved it as an idempotent function so validation was for it only. Pulling code into ServeDns.

contrib/coredns/policy/policy.go
pankaj-chhabra: Adding 
w.WriteMsg(r)

contrib/coredns/policy/policy.go
pankaj-chhabra: Removed log message

contrib/coredns/policy/policy.go
pankaj-chhabra: Removed log message

contrib/coredns/policy/policy.go
pankaj-chhabra: Removed whole section and now we just change index 0
attrs[0].Value = "response"

contrib/coredns/policy/policy.go
pankaj-chhabra: Based on the pepcli test program 
  - type: response
    policy_id: UNKNOWN
    address: 127.0.0.1

**Changed code to -**
                state := request.Request{W: w, Req: r}
		rr := r.Answer[0]
		switch state.Family() {
		case 1:
			attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: rr.(*dns.A).A})
		case 2:
			attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: rr.(*dns.AAAA).AAAA})

		}

contrib/coredns/policy/policy_test.go
pankaj-chhabra: Removed policy_test.go as we will be doing whole ServeDNS test

johnbelamaric: should still go in that file, we will just be testing the ServeDNS method of the policy middleware

contrib/coredns/policy/policy.go
johnbelamaric: You can update this PR by pushing again to your same branch.

contrib/coredns/policy/policy.go
pankaj-chhabra: I have implemented structure like this


type LocalResponseWriter struct {
	dns.ResponseWriter
	Rcode int
	Len int
	Msg   *dns.Msg
	localAddr  net.Addr
	remoteAddr net.Addr
}

func New(w dns.ResponseWriter) *LocalResponseWriter {

	var addr string = ""
	l := &net.IPAddr{IP: net.ParseIP(addr)}
	r := &net.IPAddr{IP: net.ParseIP(addr)}
	return &LocalResponseWriter{
		ResponseWriter: w,
		Rcode:          0,
		Msg:            nil,
		localAddr: l,
		remoteAddr: r,

	}
}

func (r *LocalResponseWriter) WriteMsg(res *dns.Msg) error {
	r.Rcode = res.Rcode
	// We may get called multiple times (axfr for instance).
	// Save the last message, but add the sizes.
	r.Len += res.Len()
	r.Msg = res
	return r.ResponseWriter.WriteMsg(res)
}

// Write is a wrapper that records the length of the message that gets written.
func (r *LocalResponseWriter) Write(buf []byte) (int, error) {
	n, err := r.ResponseWriter.Write(buf)
	if err == nil {
		r.Len += n
	}
	return n, err
}

my understanding is that async WriteMsg call will update the message, 

I will make the call with the new structure

localResponseWriter := new(LocalResponseWriter)
		localResponseWriter.localAddr = w.LocalAddr()
		localResponseWriter.remoteAddr = w.RemoteAddr()
		localResponseWriter.Msg = r

		upstreamResult := new(dns.Msg)

		status,err := middleware.NextOrFailure(p.Name(), p.Next, ctx, localResponseWriter, upstreamResult)
		if err != nil {
			return status, err
}

and pick the response

state := request.Request{W: localResponseWriter, Req: upstreamResult}

		rr := upstreamResult.Answer[0]
		switch state.Family() {
		case 1:
			attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: rr.(*dns.A).A.String()})
		case 2:
			attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: rr.(*dns.AAAA).AAAA.String()})

		}


I will submit Push Request with changes and will update code if my understanding is not correct.

pankaj-chhabra:
Closing the PR

PR #21, opened 2017-04-23
Response validation changes

pankaj-chhabra:
I have added changes as per the requirements I have understood, please confirm if changes are implemented correctly. Code is building, will confirm how to test it.

### DIFF
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -122,21 +122,25 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
 
 type NewLocalResponseWriter struct {
 	dns.ResponseWriter
-	Msg   *dns.Msg
+	localAddr  net.Addr
+	remoteAddr net.Addr
+	Msg        *dns.Msg
 }
 
-func (r *NewLocalResponseWriter) WriteMsg(res *dns.Msg) error {
-	r.Msg = res
-	return nil
-}
 
 func (r *NewLocalResponseWriter) Write(b []byte) (int, error) {
+	r.Msg = new(dns.Msg)
 	return len(b), r.Msg.Unpack(b)
 }
 
-// Hijack implements dns.Hijacker. It simply wraps the underlying
-// ResponseWriter's Hijack method if there is one, or returns an error.
-func (r *NewLocalResponseWriter) Hijack() { r.ResponseWriter.Hijack(); return }
+// These methods implement the dns.ResponseWriter interface from Go DNS.
+func (r *NewLocalResponseWriter) Close() error              { return nil }
+func (r *NewLocalResponseWriter) TsigStatus() error         { return nil }
+func (r *NewLocalResponseWriter) TsigTimersOnly(b bool)     { return }
+func (r *NewLocalResponseWriter) Hijack()                   { return }
+func (r *NewLocalResponseWriter) LocalAddr() net.Addr       { return r.localAddr }
+func (r *NewLocalResponseWriter) RemoteAddr() net.Addr      { return r.remoteAddr }
+func (r *NewLocalResponseWriter) WriteMsg(m *dns.Msg) error { r.Msg = m; return nil }
 
 
 func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -167,11 +167,11 @@ func (p *PolicyMiddleware) handlePermit(ctx context.Context, w dns.ResponseWrite
 	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
 
 	if err != nil {
-		return dns.RcodeRefused, nil
+		return dns.RcodeRefused, err
 	}
 
 	if !lresponse.Permit {
-		return dns.RcodeRefused, fmt.Errorf("[ERROR] Request not permitted")
+		return dns.RcodeRefused, nil
 	}
 	if lresponse.Redirect != nil {
 		return p.redirect(lresponse.Redirect.String(), lw, lw.Msg)
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -23,10 +23,6 @@ import (
 )
 
 const (
-	PolicyIdFieldName = "policy_id"
-)
-
-const (
 	EDNS0_MAP_DATA_TYPE_BYTES = iota
 	EDNS0_MAP_DATA_TYPE_HEX   = iota
 	EDNS0_MAP_DATA_TYPE_IP    = iota
@@ -62,9 +58,9 @@ type Attribute struct {
 }
 
 type Response struct {
-	Permit     bool   `pdp:"Effect"`
-	Redirect   net.IP `pdp:"redirect_to"`
-	Obligation []*Attribute
+	Permit   bool   `pdp:"Effect"`
+	Redirect net.IP `pdp:"redirect_to"`
+	PolicyId string `pdp:"policy_id"`
 }
 
 func (p *PolicyMiddleware) Connect() error {
@@ -218,12 +214,7 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 	}
 
 	if response.Permit {
-		for _, rattrs := range response.Obligation {
-			if rattrs.Id == PolicyIdFieldName {
-				attrs = append(attrs, &pb.Attribute{Id: PolicyIdFieldName, Type: "string", Value: rattrs.Value})
-				break
-			}
-		}
+		attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: response.PolicyId})
 		return p.handlePermit(ctx, w, r, attrs)
 	}
 
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -167,7 +167,7 @@ func (p *PolicyMiddleware) handlePermit(ctx context.Context, w dns.ResponseWrite
 	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
 
 	if err != nil {
-		return dns.RcodeRefused, err
+		return dns.RcodeServerFailure, err
 	}
 
 	if !lresponse.Permit {
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -51,7 +51,7 @@ type PolicyMiddleware struct {
 	ErrorFunc func(dns.ResponseWriter, *dns.Msg, int) // failover error handler
 }
 
-type  Response struct {
+type Response struct {
 	Permit   bool   `pdp:"Effect"`
 	Redirect net.IP `pdp:"redirect_to"`
 }
@@ -119,13 +119,11 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
 	return attrs, foundSourceIP
 }
 
-
 type NewLocalResponseWriter struct {
 	localAddr  net.Addr
 	remoteAddr net.Addr
 	Msg        *dns.Msg
 }
-
 
 func (r *NewLocalResponseWriter) Write(b []byte) (int, error) {
 	r.Msg = new(dns.Msg)
@@ -140,7 +138,6 @@ func (r *NewLocalResponseWriter) Hijack()                   { return }
 func (r *NewLocalResponseWriter) LocalAddr() net.Addr       { return r.localAddr }
 func (r *NewLocalResponseWriter) RemoteAddr() net.Addr      { return r.remoteAddr }
 func (r *NewLocalResponseWriter) WriteMsg(m *dns.Msg) error { r.Msg = m; return nil }
-
 
 func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {
 	state := request.Request{W: w, Req: r}
@@ -172,9 +169,9 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 		return dns.RcodeServerFailure, err
 	}
 
-	if response.Permit == true {
+	if response.Permit {
 		lw := new(NewLocalResponseWriter)
-		status,err := middleware.NextOrFailure(p.Name(), p.Next, ctx, lw, r)
+		status, err := middleware.NextOrFailure(p.Name(), p.Next, ctx, lw, r)
 		if err != nil {
 			return status, err
 		}
@@ -191,7 +188,7 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 			default:
 			}
 		}
-		attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: ""})
+		attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: "SECURITY_POLICY_1790"})
 
 		var lresponse Response
 
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -23,6 +23,10 @@ import (
 )
 
 const (
+	PolicyIdFieldName = "policy_id"
+)
+
+const (
 	EDNS0_MAP_DATA_TYPE_BYTES = iota
 	EDNS0_MAP_DATA_TYPE_HEX   = iota
 	EDNS0_MAP_DATA_TYPE_IP    = iota
@@ -51,9 +55,16 @@ type PolicyMiddleware struct {
 	ErrorFunc func(dns.ResponseWriter, *dns.Msg, int) // failover error handler
 }
 
+type Attribute struct {
+	Id    string
+	Type  string
+	Value string
+}
+
 type Response struct {
-	Permit   bool   `pdp:"Effect"`
-	Redirect net.IP `pdp:"redirect_to"`
+	Permit     bool   `pdp:"Effect"`
+	Redirect   net.IP `pdp:"redirect_to"`
+	Obligation []*Attribute
 }
 
 func (p *PolicyMiddleware) Connect() error {
@@ -142,7 +153,7 @@ func (r *NewLocalResponseWriter) WriteMsg(m *dns.Msg) error { r.Msg = m; return 
 // This function validates the response from middleware
 // Inputs are - context.Context, dns.ResponseWriter, *dns.Msg, policy_id, attrs
 // Outputs are - int, error
-func (p *PolicyMiddleware) handlePermit(ctx context.Context, w dns.ResponseWriter, r *dns.Msg, policy_id string, attrs []*pb.Attribute) (int, error) {
+func (p *PolicyMiddleware) handlePermit(ctx context.Context, w dns.ResponseWriter, r *dns.Msg, attrs []*pb.Attribute) (int, error) {
 	lw := new(NewLocalResponseWriter)
 	status, err := middleware.NextOrFailure(p.Name(), p.Next, ctx, lw, r)
 	if err != nil {
@@ -162,7 +173,6 @@ func (p *PolicyMiddleware) handlePermit(ctx context.Context, w dns.ResponseWrite
 		}
 	}
 
-	attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: policy_id})
 	var lresponse Response
 	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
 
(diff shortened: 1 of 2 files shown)


contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+type LocalResponseWriter struct {
+	dns.ResponseWriter
+	Rcode int
+	Len int
+	Msg   *dns.Msg
+	localAddr  net.Addr
+	remoteAddr net.Addr
+}
+
+func New(w dns.ResponseWriter) *LocalResponseWriter {
```
johnbelamaric: This should be "NewLocalResponseWriter".

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+
+	var addr string = ""
+	l := &net.IPAddr{IP: net.ParseIP(addr)}
+	r := &net.IPAddr{IP: net.ParseIP(addr)}
+	return &LocalResponseWriter{
+		ResponseWriter: w,
+		Rcode:          0,
```
johnbelamaric: this is the default, you can leave out this line

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+		Msg:            nil,
```
johnbelamaric: ditto

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
 	return attrs, foundSourceIP
 }
 
+
+type LocalResponseWriter struct {
+	dns.ResponseWriter
+	Rcode int
+	Len int
+	Msg   *dns.Msg
```
johnbelamaric: do we need Rcode and Len or just Msg?

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +196,54 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 	}
 
 	if response.Permit {
-		return middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
+
+		lr := new(LocalResponseWriter)
```
johnbelamaric: Use NewLocalResponseWriter instead and then you can get rid of the three lines that follow this.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
 	return attrs, foundSourceIP
 }
 
+
+type LocalResponseWriter struct {
+	dns.ResponseWriter
```
johnbelamaric: I don't think we actually need to keep the response writer in here. All this LocalResponseWriter is trying to do is capture any response locally - not to write it back to the client.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+	}
+}
+
+func (r *LocalResponseWriter) WriteMsg(res *dns.Msg) error {
+	r.Rcode = res.Rcode
+	// We may get called multiple times (axfr for instance).
+	// Save the last message, but add the sizes.
+	r.Len += res.Len()
+	r.Msg = res
+	return r.ResponseWriter.WriteMsg(res)
```
johnbelamaric: Actually, we do NOT want to do the WriteMsg here. The point is we want to just capture the `res` in r.Msg so we can do the WriteMsg *later* if we decide to keep it.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,52 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+}
+
+// Write is a wrapper that records the length of the message that gets written.
+func (r *LocalResponseWriter) Write(buf []byte) (int, error) {
+	n, err := r.ResponseWriter.Write(buf)
```
johnbelamaric: Similar to above - don't actually Write here. See how the grpc one does it - it converts the Write into a Msg.

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +196,54 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
-		return middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
+
+		lr := new(LocalResponseWriter)
+		lr.localAddr = w.LocalAddr()
+		lr.remoteAddr = w.RemoteAddr()
+		lr.Msg = r
+
+		upstreamResult := new(dns.Msg)
+		upstreamResult = r
+		status,err := middleware.NextOrFailure(p.Name(), p.Next, ctx, lr, upstreamResult)
```
johnbelamaric: `upstreamResult` is not needed. You can just pass `r`. The critical thing is how you have replaced `w` with `lr`.

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +196,54 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+		if err != nil {
+			return status, err
+		}
+
+		state := request.Request{W: lr, Req: upstreamResult}
+
+		rr := upstreamResult.Answer[0]
```
johnbelamaric: I think where you want to look is in lr.Msg, not upstreamResult.

contrib/coredns/policy/policy.go
pankaj-chhabra: dns.ResponseWriter interface is required as NextOfFailure requires dns.ResponseWriter parameter

func NextOrFailure(name string, next Handler, ctx context.Context, w dns.ResponseWriter, r *dns.Msg) (int, error) {

johnbelamaric: You don't need to include it here in order for the struct to implement the interface. Just having the right methods defined does it.

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+		status,err := middleware.NextOrFailure(p.Name(), p.Next, ctx, lr, r)
+		if err != nil {
+			return status, err
+		}
+
+		state := request.Request{W: lr, Req: lr.Msg}
+
+		rr := lr.Msg.Answer[0]
+		switch state.Family() {
+		case 1:
```
johnbelamaric: Is there a constant defined for these values? Maybe in dns?

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+			attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: rr.(*dns.A).A.String()})
```
johnbelamaric: Does this type cast work here? Probably you should do a type assertion using the value, ok idiom. In that case you could get rid of the `state` variable since you won't need to check `Family()`. Something like:

```
if a, ok := rr.(*dns.A); ok {
    attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: a.A.String()})
} else if a, ok := rr.(*dns.AAAA); ok {
    attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: a.AAAA.String()})
}
```

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+		case 2:
+			attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: rr.(*dns.AAAA).AAAA.String()})
+
+		}
+
+		attrs[0].Value = "response"
+
+		var response Response
```
johnbelamaric: Doesn't `response` already exist? I guess this is nested in the `Permit` scope. Best to use a different name.

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+		err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &response)
```
johnbelamaric: We also will need to copy the `policy_id` attribute out of the original response obligations.

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+
+		if err != nil {
+			log.Printf("[ERROR] Policy validation failed due to error %s\n", err)
+			return dns.RcodeServerFailure, err
+		}
+
+		if response.Permit == true && response.Redirect == nil {
+			w.WriteMsg(r)
```
johnbelamaric: shouldn't this be `lr.Msg()` instead of `r`?

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
...
+			return status, nil
+		}
+
+		if response.Redirect != nil {
+			log.Printf("[INFO] Redirecting request")
+			return p.redirect(response.Redirect.String(), w, r)
+		}
+
+		if response.Permit == false {
```
johnbelamaric: This `if` is not needed. We don't need to log this - just let it go through to the return on line 217.

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +174,47 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 	}
 
 	if response.Permit {
```
johnbelamaric: If you want to create a separate function for this stuff for unit testing purposes (as you did before), then I think it's OK, but you should call it something like `handlePermit`, and it should take the same params as `ServeDNS`, plus the policy response.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,70 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+		default:
+		}
+	}
+
+	// Dummy security policy
+	attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: "SECURITY_POLICY_1790"})
+	var lresponse Response
+	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
+
+	if err == nil {
```
johnbelamaric: Change to != and return refused, err. Then you don't have to beat the other things.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,70 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+		if lresponse.Permit == false {
+			return dns.RcodeRefused, fmt.Errorf("[ERROR] Request not permitted")
+		} else {
```
johnbelamaric: Don't need "else" since the above returns.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,70 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+		}
+	}
+
+	// Dummy security policy
+	attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: "SECURITY_POLICY_1790"})
+	var lresponse Response
+	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
+
+	if err == nil {
+		if lresponse.Permit == false {
```
johnbelamaric: If !lresponse.Permit {

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,70 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+				attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: a.A.String()})
+			} else if a, ok := rr.(*dns.AAAA); ok {
+				attrs = append(attrs, &pb.Attribute{Id: "address", Type: "address", Value: a.AAAA.String()})
+			}
+		default:
+		}
+	}
+
+	// Dummy security policy
+	attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: "SECURITY_POLICY_1790"})
```
johnbelamaric: You'll need to pass in the original Response as well, to get the policy ID. Or at least the policy ID itself.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,70 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+	var lresponse Response
+	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
+
+	if err == nil {
+		if lresponse.Permit == false {
+			return dns.RcodeRefused, fmt.Errorf("[ERROR] Request not permitted")
+		} else {
+			if lresponse.Redirect != nil {
+				return p.redirect(lresponse.Redirect.String(), lw, lw.Msg)
+			} else {
```
johnbelamaric: Don't need "else"

contrib/coredns/policy/policy_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/miekg/dns"
+	"testing"
+)
+
+// This function unit tests handlePermit function
+// Inputs are t *testing.T
+// Output is update in t
+func TestHandlePermit(t *testing.T) {
+	p := new(PolicyMiddleware)
+	p.pdp = new(pep.Client)
```
johnbelamaric: We'll need to make Client an interface and create a testClient.

pankaj-chhabra: For now I have removed the Client, which is not required for this Unit test. For next Unit test, will create a Client and create a test Client

contrib/coredns/policy/policy.go
pankaj-chhabra: Changed it

contrib/coredns/policy/policy.go
pankaj-chhabra: Changed it

contrib/coredns/policy/policy.go
pankaj-chhabra: Changed it

contrib/coredns/policy/policy.go
pankaj-chhabra: Changed it

contrib/coredns/policy/policy.go
pankaj-chhabra: Added policy ID

contrib/coredns/policy/policy.go
```diff
@@ -162,25 +162,22 @@ func (p *PolicyMiddleware) handlePermit(ctx context.Context, w dns.ResponseWrite
...
-		} else {
-			if lresponse.Redirect != nil {
-				return p.redirect(lresponse.Redirect.String(), lw, lw.Msg)
-			} else {
-				w.WriteMsg(lw.Msg)
-				return status, nil
-			}
-		}
+	if err != nil {
+		return dns.RcodeRefused, nil
```
johnbelamaric: Return err also, not nil

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,67 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+	attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: policy_id})
+	var lresponse Response
+	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
+
+	if err != nil {
+		return dns.RcodeRefused, nil
+	}
+
+	if !lresponse.Permit {
+		return dns.RcodeRefused, fmt.Errorf("[ERROR] Request not permitted")
```
johnbelamaric: Return nil, not fmt.Errorf here; this isn't an error.

contrib/coredns/policy/policy.go
pankaj-chhabra: Done

contrib/coredns/policy/policy.go
pankaj-chhabra: Done

pankaj-chhabra: Please confirm if the status is correct as well, I have code it RcodeRefused, original code returns dns.RcodeServerFailure. I think it should be RcodeRefused. Please confirm

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,67 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+		default:
+		}
+	}
+
+	attrs = append(attrs, &pb.Attribute{Id: "policy_id", Type: "string", Value: policy_id})
+	var lresponse Response
+	err = p.pdp.Validate(ctx, pb.Request{Attributes: attrs}, &lresponse)
+
+	if err != nil {
+		return dns.RcodeRefused, err
```
johnbelamaric: Should be dns.RcodeServerFailure as you suggested.

contrib/coredns/policy/policy.go
```diff
@@ -119,6 +119,67 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
...
+	}
+
+	if !lresponse.Permit {
+		return dns.RcodeRefused, nil
```
johnbelamaric: Refused is correct here.

contrib/coredns/policy/policy.go
pankaj-chhabra: Changed

contrib/coredns/policy/policy.go
```diff
@@ -51,9 +55,16 @@ type PolicyMiddleware struct {
...
+	Type  string
+	Value string
+}
+
 type Response struct {
-	Permit   bool   `pdp:"Effect"`
-	Redirect net.IP `pdp:"redirect_to"`
+	Permit     bool   `pdp:"Effect"`
+	Redirect   net.IP `pdp:"redirect_to"`
+	Obligation []*Attribute
```
johnbelamaric: I think you can just add

```
 PolicyId string `pdp:"policy_id"`
```

here? @vasili-v ?

contrib/coredns/policy/policy.go
```diff
@@ -150,7 +218,13 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 	}
 
 	if response.Permit {
-		return middleware.NextOrFailure(p.Name(), p.Next, ctx, w, r)
+		for _, rattrs := range response.Obligation {
+			if rattrs.Id == PolicyIdFieldName {
```
johnbelamaric: If I am right about the above comment, then this just becomes response.PolicyId.

contrib/coredns/policy/policy.go
vasili-v: Yes. It should look like @johnbelamaric wrote above. You don't need Obligation array just this PolicyId field

pankaj-chhabra: Hello, my understanding is that these changes are checked in, I could not find policy_id in PDP server. In service.pb.response, Response structure is 

type Response struct {
	Effect     Response_Effect `protobuf:"varint,1,opt,name=effect,enum=service.Response_Effect" json:"effect,omitempty"`
	Reason     string          `protobuf:"bytes,2,opt,name=reason" json:"reason,omitempty"`
	Obligation []*Attribute    `protobuf:"bytes,3,rep,name=obligation" json:"obligation,omitempty"`
}

We make the validate call 

func (c *pDPClient) Validate(ctx context.Context, in *Request, opts ...grpc.CallOption) (*Response, error) {
	out := new(Response)
	err := grpc.Invoke(ctx, "/service.PDP/Validate", in, out, c.cc, opts...)

Returned structure is Response above.

I am making the change assuming that fillResponse in client will return PolicyId

contrib/coredns/policy/policy_test.go
```diff
@@ -0,0 +1,28 @@
...
+func TestHandlePermit(t *testing.T) {
+	p := new(PolicyMiddleware)
+	tw := new(NewLocalResponseWriter)
+	ctx := context.Background()
+
+	r := new(dns.Msg)
+	r.SetQuestion("example.com.", dns.TypeANY)
+	attrs := []*pb.Attribute{&pb.Attribute{Id: "type", Type: "string", Value: "query"}}
+
+	status, err := p.handlePermit(ctx, tw, r, "SECURITY_POLICY_1790", attrs)
```
johnbelamaric: As written, this won't work. You need to add "contrib/coredns/policy" to the TESTABLE variable here:

https://github.com/infobloxopen/themis/blob/master/Makefile#L2

so that this is run when the CI runs.

yuewko: Also, you might need to do a rebase in order to get around the codecov/project failure.

PR #28, opened 2017-05-03
fix import in the test

mstaneuka:


contrib/coredns/policy/policy_test.go
```diff
@@ -2,7 +2,7 @@ package policy
 
 import (
 	"context"
-	pb "github.com/infobloxopen/policy-box/pdp-service"
+	pb "github.com/infobloxopen/themis/pdp-service"
```
mstaneuka: glide doesn't recognize repo forward so this fix removes policy-box from vendor as well

PR #32, opened 2017-05-31
Add support of uploading versioned data to the pdpctrl-client

mstaneuka:


proto/control.proto
```diff
@@ -11,12 +11,15 @@ service PDPControl {
...
+    CONTENT = 1;
+    POLICIES_PATCH = 2;
+    CONTENT_PATCH = 3;
   }
-           DataType type     = 1;
-           int32    dataId   = 2;
-           string   id       = 3;
-  repeated int32    includes = 4;
+  DataType type = 1;
+  string version = 2;
```
vasili-v: I see here only one version field and now think that it is version of PDP's data we try update. However it's not enough we also need to tell PDP to which version this update switches PDP.

pdpctrl-client/host.go
```diff
@@ -11,57 +13,88 @@ import (
...
+func (h *Host) Close() {
 	if h.conn != nil {
 		h.conn.Close()
 	}
 
 	h.client = nil
 }
 
-func (h *Host) uploadIncludes(includes map[string][]byte, log Logger) []int32 {
+func (h *Host) Process(patch bool, version string, includes map[string][]byte, policy []byte) error {
```
mstaneuka: >The main part of cluster abstraction was to separate upload and apply into two phases on whole set of servers

Why do we need this property? If some uploads will fail we end up with partially updated PDPs anyway. From the other hand independent PDP management should simplify our logic

proto/control.proto
mstaneuka: This is actually ```to witch version```. It's assumed that PAP tracks current version for each PDP and creates patch from this version to latest one and sends only latest one. In order to validate on PDP side it's perhaps make sense to send both version - from and to...

pdpctrl-client/host.go
```diff
@@ -70,197 +74,136 @@ func (h *Host) Close() {
 	h.client = nil
 }
 
-func (h *Host) Process(patch bool, version string, includes map[string][]byte, policy []byte) error {
-	if err := h.upload(patch, version, includes, policy); err != nil {
+func (h *Host) Upload(bucket *DataBucket) error {
+	h.log.Infof("Uploading PDP data to host %s...", h.address)
```
vasili-v: We don't need logger any more. Previously error handling was incapsulated to cluster and outer code passes logger to get notion on errors. Now all functions return errors and caller can do whatever it want.

papcli/main.go
```diff
@@ -5,26 +5,57 @@ import (
...
+
+	bids := make([]int32, len(hosts))
+	for i, h := range hosts {
+		b := &pdpctrl.DataBucket{
+			Policies: policies,
+			Includes: includes,
+		}
+
+		if err := h.Upload(b); err != nil {
+			log.Errorf("Failed to upload data: %v", err)
```
johnbelamaric: If we get a version error, shouldn't we automatically push the entire policy? It's not clear to me that we are doing that.

mstaneuka: This should be responsibility of client's user (e.g. PAP implementation) because we need to recreate binary content and upload new data bucket.

Also version error contains correct version number so we not necessary need entire copy

PR #33, opened 2017-06-13
PDP patched update

mstaneuka:


mstaneuka:
Need more UT cases and integration testing.

pdpserver/patch_test.go
```diff
@@ -0,0 +1,255 @@
...
+            effect: Permit
+      - id: P3P2
+        alg: FirstApplicableEffect
+        rules:
+          - id: P3P2R1
+            effect: Permit
+`
+
+func TestSelectorUpdatesOnContentPatches(t *testing.T) {
+	ctx := pdp.NewYASTCtx("")
```
mstaneuka: TBD: going to add many more UT cases

pdp/mapper_rca.go
```diff
@@ -34,33 +34,33 @@ func getSetOfIDs(v AttributeValueType) ([]string, error) {
...
 
-func getRulesMap(rules []RuleType, params *MapperRCAParams) map[string]*RuleType {
+func getRulesMap(rules []*RuleType, params *MapperRCAParams) map[string]*RuleType {
 	if params.RulesMap != nil {
 		return params.RulesMap
 	}
 
 	m := make(map[string]*RuleType)
 	for _, rule := range rules {
 		tmp := rule
```
vasili-v: In case of pointers we don't need `tmp` variable. You need to put pointer directly to the list.

pdp/yast_policies.go
```diff
@@ -79,7 +79,7 @@ func unmarshalMapperRCAParams(ctx *YastCtx, p *PolicyType, root bool, m map[inte
 	rulesMap := make(map[string]*RuleType)
 	for _, r := range p.Rules {
 		tmp := r
```
vasili-v: The same as before. Please remove `tmp`.

mstaneuka:
This PR is finished - I'm not going to add more changes except for addressing any new comments.
For policy patches Copy-On-Write mechanism is implemented.
Content is update "in place" since it's not used directly by policy, this introduce another issue - some policies might need to be updated (alg selectors) if content has been patched. This issue also addressed by introducing "internal" refresh command - set of such commands is generated during content patching.

All logic assumes that in case of any error during patching PAP must send full data next time, but policies will never damaged.

PR #35, opened 2017-06-14
Added 32/64-bit radix tree with Insert and Get operations

vasili-v:
John, when you create new repo I move the code to it


johnbelamaric:
Ok, what else are you planning to put in the repo...more to the point what repo name do you want?

PR #36, opened 2017-06-16
Turn ON domain validation

UladzimirTrehubenka:


pdp/domain_tree.go
```diff
@@ -18,8 +19,10 @@ type DomainLeafItem struct {
 	Leaf   interface{}
 }
 
+var globalIdnaInstance = idna.New(idna.MapForLookup())
```
vasili-v: idna library provides 4 validation profiles:
* Punycode is a Profile that does raw punycode processing with a minimum of validation.
* Lookup is the recommended profile for looking up domain names, according to Section 5 of RFC 5891. The exact configuration of this profile may change over time.
* Display is the recommended profile for displaying domain names. The configuration of this profile may change over time.
* Registration is the recommended profile for checking whether a given IDN is valid for registration, according to Section 4 of RFC 5891.

By default the library uses "Punycode" profile. And we need to make agreement for entire system (portal, janus and policy engine) before switching to any particular level.

PR #38, opened 2017-06-21
Fix content index update on policy add/replace

mstaneuka:


vasili-v:
I approve the request but we need some test coverage

PR #41, opened 2017-07-11
Improved domain validation, apend 'customer_id' attribute handling

UladzimirTrehubenka:


contrib/coredns/policy/policy.go
```diff
@@ -199,6 +198,23 @@ func (p *PolicyMiddleware) ServeDNS(ctx context.Context, w dns.ResponseWriter, r
 
 	edns, foundSourceIP := p.getEDNS0Attrs(r)
 	if len(edns) > 0 {
+		clientIDok := false
+		customerIDok := false
+		customerID := ""
+		for _, entry := range edns {
```
johnbelamaric: This code should not know anything about specific attributes. It is general purpose policy code. You need to find another way to do this. Why wouldn't these both just be configured in the originating client?

PR #43, opened 2017-07-11
Improved handling EDNS attributes

UladzimirTrehubenka:
Added additional params for string conversion - stringOffset and stringSize.
Refactor EDNS attributes handling - set empty string ("") if we have attribute in config but don't have it in EDNS.

Example Corefile:
	policy {
		endpoint 127.0.0.1:5555
                # edns0 0xffee client_id
		edns0 0xffee customer_id hex string 0 32 # make customer_id from first 16 bytes of client_id
		# edns0 0xfff1 customer_id
		edns0 0xffef policy_id
		edns0 0xfff2 on_prem_host_id
	}

### DIFF
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -85,11 +85,11 @@ func (p *PolicyMiddleware) AddEDNS0Map(code, name, dataType, destType,
 	}
 	offset, err := strconv.Atoi(stringOffset)
 	if err != nil {
-		return fmt.Errorf("Could not parse EDNS0 code: %s", err)
+		return fmt.Errorf("Could not parse EDNS0 string offset: %s", err)
 	}
 	size, err := strconv.Atoi(stringSize)
 	if err != nil {
-		return fmt.Errorf("Could not parse EDNS0 code: %s", err)
+		return fmt.Errorf("Could not parse EDNS0 string size: %s", err)
 	}
 	ednsType, ok := stringToEDNS0MapType[dataType]
 	if !ok {
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -35,10 +35,12 @@ var stringToEDNS0MapType = map[string]uint16{
 }
 
 type edns0Map struct {
-	code     uint16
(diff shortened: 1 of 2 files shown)


contrib/coredns/policy/policy.go
```diff
@@ -75,16 +77,25 @@ func (p *PolicyMiddleware) Connect() error {
...
-func (p *PolicyMiddleware) AddEDNS0Map(code, name, dataType, destType string) error {
+func (p *PolicyMiddleware) AddEDNS0Map(code, name, dataType, destType,
+	stringOffset, stringSize string) error {
 	c, err := strconv.ParseUint(code, 0, 16)
 	if err != nil {
 		return fmt.Errorf("Could not parse EDNS0 code: %s", err)
 	}
+	offset, err := strconv.Atoi(stringOffset)
+	if err != nil {
+		return fmt.Errorf("Could not parse EDNS0 code: %s", err)
```
johnbelamaric: s/code/string offset/

contrib/coredns/policy/policy.go
```diff
@@ -75,16 +77,25 @@ func (p *PolicyMiddleware) Connect() error {
...
+	}
+	size, err := strconv.Atoi(stringSize)
+	if err != nil {
+		return fmt.Errorf("Could not parse EDNS0 code: %s", err)
```
johnbelamaric: s/code/string size/

contrib/coredns/policy/policy.go
```diff
@@ -116,12 +128,17 @@ func (p *PolicyMiddleware) getEDNS0Attrs(r *dns.Msg) ([]*pb.Attribute, bool) {
 						ip := net.IP(e.Data)
 						value = ip.String()
 					}
+					from := m.stringOffset
+					to := m.stringOffset + m.stringSize
```
johnbelamaric: If Offset and Size are not specified, they are both "0", isn't that an issue - or did I miss something?

UladzimirTrehubenka: Not an issue - in this case (0/0) string is not truncated

PR #45, opened 2017-07-31
Themis reassembly

vasili-v:
This is preliminary review. Shouldn't be merged. Too breaking... Additionally need to restore test coverage and fix related code.


johnbelamaric:
Can we go through this Thursday or Friday? I am on PTO the next two weeks after that. I need a walkthrough...

PR #56, opened 2017-09-01
Fixed processing requests when there is no any policy and enhanced pep

vasili-v:


vasili-v:
Unfortunately, it's quite not possible to increase test coverage.

PR #62, opened 2017-09-21
Added debug query support

UladzimirTrehubenka:


johnbelamaric:
Ok. LGTM. I think it's time to add a README file to this. Take a look at the other plugin README files in https://github.com/coredns/coredns and model it after those.

PR #65, opened 2017-09-25
Add support for dnstap PolicyHit message

rdrozhdzh:
- policy plugin generates and sends PolicyHit message with
   results of PDP server desision
 - PolicyHit messages are sent only when dnstap plugin is enabled


contrib/coredns/policy/policytap/policytap_test.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil, err
+	}
+	phMsg := &PolicyHitMessage{}
+	ext, err := proto.GetExtension(dMsg, E_PolicyHit)
+	if err == nil {
+		phMsg = ext.(*PolicyHitMessage)
+	} else {
+		//Dnstap message is not extendable? Checking XXX_unrecognized
+		if dMsg.XXX_unrecognized != nil && len(dMsg.XXX_unrecognized) > 0 {
+			wraper := &PolicyHitMessageWraper{}
```
johnbelamaric: s/wraper/wrapper/

contrib/coredns/policy/policytap/policytap_msg.go
```diff
@@ -0,0 +1,84 @@
...
+
+	dnstap "github.com/dnstap/golang-dnstap"
+	"github.com/golang/protobuf/proto"
+	pb "github.com/infobloxopen/themis/pdp-service"
+	"github.com/miekg/dns"
+)
+
+// ToDnstap creates PolicyHitMessage and writes it to the privided Writer
+func ToDnstap(w io.Writer, t time.Time, msg *dns.Msg, tt PolicyHitMessage_PolicyTriggerType,
+	attrs []*DnstapAttribute, r *pb.Response) error {
```
johnbelamaric: It may be a good idea to spin off another go routine for this. So that the primary query doesn't have to wait for the DNS tap message to be sent.

rdrozhdzh: Ok, will do

johnbelamaric: Also, it should be checking if the Send CR flag is set as per https://github.com/coredns/coredns/pull/1101

rdrozhdzh: Hm, not sure if CR flag should be checked for PolicyHit messages

johnbelamaric: Oops. Nevermind. Confused myself.

contrib/coredns/policy/policytap/policytap_test.go
rdrozhdzh: done

contrib/coredns/policy/policytap/policytap_msg.go
rdrozhdzh: Done sending in a separate go routine

rdrozhdzh:
Refactor PolicyHit message implementation
 - Pass dnstap message to recently introduced IORoutine interface
 - Rename message PolicyHitMessageWraper to PolicyHitMessageWrapper
 - Send PolicyHit message in a separate go routine
 - Fix GoVet and GoLint warnings
 - Other minor changes

contrib/coredns/policy/policytap/policytap.proto
```diff
@@ -0,0 +1,72 @@
...
+	required uint64 time_sec = 1;
+	required fixed32 time_nsec = 2;
+
+	// DNS query/transaction id; this is a 16-bit value, but
+	// ProtocolBuffers interface defines only longer integers.
+	required uint32 query_id = 3;
+
+	// Policy Trigger Type -- domain name or address
+	required PolicyTriggerType trigger_type = 4;
+
```
huynhmb359: Where are the trigger_name and trigger_category?  I think PDP will be able to provide these 2 info.  We should define them here as well.

rdrozhdzh: trigger_name and trigger_category should be configured in PAP. If configured, PDP will be return these attributes as obligations, and they will be copied to PolicyHitMessage.attributes

PR #66, opened 2017-09-26
Fixed lost source_ip if no EDNS0 options

UladzimirTrehubenka:


contrib/coredns/policy/policy.go
```diff
@@ -144,37 +144,35 @@ func (p *PolicyPlugin) getAttrsFromEDNS0(r *dns.Msg, ip string) []*pb.Attribute
 	var attrs []*pb.Attribute
 
 	o := r.IsEdns0()
-	if o == nil {
```
johnbelamaric: Why did you remove this and nest the for loop? It makes it harder to read due to more indent levels. Also harder to see what you really changed.

UladzimirTrehubenka: Otherwise we need move foundSourceIP logic outside getAttrsFromEDNS0()
or duplicate it.

PR #70, opened 2017-09-29
Added memory check and GC control

vasili-v:


vasili-v:
Later if we found that performance essentially degrades with the 10% of GC we can tune it and add automation depending on memory pressure

johnbelamaric:
Looks like a go fmt failure.

PR #71, opened 2017-10-02
Improved debug query handling

UladzimirTrehubenka:


contrib/coredns/policy/policy.go
```diff
@@ -269,6 +269,7 @@ func (p *PolicyPlugin) handlePermit(ctx context.Context, w dns.ResponseWriter,
 	resp := makeResponse(response)
 
 	if debugQuery && resp.Action != typeRefuse {
+		response.Obligation = append(response.Obligation, respDomain.Obligation...)
```
vasili-v: Don't see why you want to join them. Just pass two responses to the function and format them properly.

PR #73, opened 2017-10-04
Improved edns0 options handling

UladzimirTrehubenka:
ATM we have support for splitting any edns0 option to some separate attributes,
syntax:  edns0 [code]  [name] [dataType] [destType] [start] [end]
where [dataType] = hex, [start] is the start index in byte array, [end] is the end index in byte array + 1
Example:
edns0 0xffee customer_id hex string 0 16
edns0 0xffee client_id hex string 16 32
Attributes (data has size 32 bytes - getting from the option with 0xffee code):
customer_id = data[0:16] (first 16 bytes)
client_id = data[16:32] (last 16 bytes)

### DIFF
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -157,6 +157,36 @@ func (p *PolicyPlugin) AddEDNS0Map(code, name, dataType, destType,
 	return nil
 }
 
+func parseOptionGroup(data []byte, options []edns0Map) ([]*pb.Attribute, bool) {
+	srcIpFound := false
+	var attrs []*pb.Attribute
+	for _, option := range options {
+		var value string
+		switch option.dataType {
+		case typeEDNS0Bytes:
+			value = string(data)
+		case typeEDNS0Hex:
+			start := uint(0)
+			end := uint(len(data))
+			if option.start < end {
+				start = option.start
+			}
+			if option.end < end && option.end > 0 {
+				end = option.end
+			}
+			value = hex.EncodeToString(data[start:end])
+		case typeEDNS0IP:
+			ip := net.IP(data)
+			value = ip.String()
+		}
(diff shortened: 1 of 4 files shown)


contrib/coredns/policy/policy.go
```diff
@@ -245,7 +250,7 @@ func (p *PolicyPlugin) handlePermit(ctx context.Context, w dns.ResponseWriter,
 		if debugQuery {
 			return p.retDebugInfo(r, w, respDomain, nil, "failed")
 		}
-		return status, err
+		return p.retRcode(w, r, dns.RcodeServerFailure, err)
```
rdrozhdzh: Why did you change status to dns.RcodeServerFailure?

UladzimirTrehubenka: If err != nil status is RcodeServerFailure, so the logic is not broken.

UladzimirTrehubenka: status is used later in the code

contrib/coredns/policy/policy.go
```diff
@@ -171,29 +171,34 @@ func (p *PolicyPlugin) getAttrsFromEDNS0(r *dns.Msg, ip string) []*pb.Attribute
...
-			value = ip.String()
-		}
-		from := option.stringOffset
-		to := option.stringOffset + option.stringSize
-		if to > 0 && to <= len(value) && from < to {
-			value = value[from:to]
-		}
-		if option.name == "source_ip" {
-			ipId = "proxy_source_ip"
+		for _, option := range options {
```
vasili-v: May be extract this to separate function to put following here:
```go
...
group, srcIp := parseOptionGroup(options)
attrs = append(attrs, group...)
if srcIp {
	ipId = "proxy_source_ip"
}
...
```

PR #74, opened 2017-10-04
Adding new gRPC call to notify PDP about its readiness

skudriavtsev:


proto/control.proto
```diff
@@ -6,6 +6,7 @@ service PDPControl {
   rpc Request (Item) returns (Response) {}
   rpc Upload (stream Chunk) returns (Response) {}
   rpc Apply (Update) returns (Response) {}
+  rpc NotifyReady (Empty) returns (Response) {}
```
vasili-v: Don't gRPC allow method with no arguments?

pdpserver/control.go
```diff
@@ -164,3 +164,10 @@ func (s *server) Apply(ctx context.Context, in *pb.Update) (*pb.Response, error)
 
 	return res, err
 }
+
+func (s *server) NotifyReady(ctx context.Context, m *pb.Empty) (*pb.Response, error) {
+	log.Info("Got notified about readiness")
+	s.start<-true
```
vasili-v: Instead of stopping there and then resuming here it's better to make serveRequests function and just call it here

proto/control.proto
skudriavtsev: With empty brackets there will be an error. Thus we do have an argument: Empty (it is a message).

pdpserver/control.go
```diff
@@ -164,3 +164,10 @@ func (s *server) Apply(ctx context.Context, in *pb.Update) (*pb.Response, error)
 
 	return res, err
 }
+
+func (s *server) NotifyReady(ctx context.Context, m *pb.Empty) (*pb.Response, error) {
+	log.Info("Got notified about readiness")
+	s.start <- true
```
vasili-v: Instead of stopping there and then resuming here it's better to make serveRequests function and just call it here.

skudriavtsev: Actually it is not so easy, because this function will require a tracer parameter to be passed (as the code bellow needs it). But from NotifyReady handler we have no access to it. Such approach will require some (probably not so optimal) refactoring. Global variable or additional field of the 'server' structure. I think, it is not a good idea.

vasili-v: You're right. Tracer the same as interface and protocol should be landed to server struct. Issue with current code that second call to NotifyReady if happens accidentally hangs forever. There is non-blocking channel write to deal with it:
```go
select {
default:
case s.start <- true:
}
```
But I'd prefer to change s.start to s.tracer and call serveRequests here instead to be more explicit.

skudriavtsev:
> we should also add a readiness check

I think it will be the next step. Even now we can perform an implicit healthcheck just trying to connect to tcp/5555

pdpserver/control.go
```diff
@@ -164,3 +164,15 @@ func (s *server) Apply(ctx context.Context, in *pb.Update) (*pb.Response, error)
...
 	return res, err
 }
+
+func (s *server) NotifyReady(ctx context.Context, m *pb.Empty) (*pb.Response, error) {
+	log.Info("Got notified about readiness")
+
+	// PDP server needs 'ready' signal from PAP server only if there
+	// are no initialization policy and content files provided.
+	if len(conf.policy) == 0 && len(conf.content) == 0 {
+		s.start <- true
```
vasili-v: Still stops forever on second call.

pdpserver/server.go
```diff
@@ -241,4 +241,7 @@ func (s *server) serve() {
 		// serveRequests() will be executed by external request.
 		log.Info("Waiting for policies to be applied.")
 	}
+	for {
+		time.Sleep(time.Second)
```
vasili-v: Periodic sleeping is not a good way to wait forever. There are [bunch of better ways to do it](https://blog.sgmansfield.com/2016/06/how-to-block-forever-in-go/)

PR #75, opened 2017-10-05
Send dnstap CR message with extra attributes from policy plugin

rdrozhdzh:


contrib/coredns/policy/policy_test.go
```diff
@@ -467,3 +471,60 @@ func TestEdns(t *testing.T) {
...
+	permit := &pdp.Response{Effect: pdp.Response_PERMIT}
+	pm.pdp = newTestClientInit(permit, permit, nil, nil)
+
+	trapper := dtest.TrapTapper{Full: true}
+	tapRW := taprw.ResponseWriter{
+		Query:          new(dns.Msg),
+		ResponseWriter: &test.ResponseWriter{},
+		Tapper:         &trapper,
+	}
+	pm.ServeDNS(context.TODO(), &tapRW, req)
```
vasili-v: I prefer to have here context.Background because don't expect that you or anyone is going to change it to something else.

vasili-v:
Actually, I didn't get logic of the new code. Need to be familiar with DNS tap to review it.

contrib/coredns/policy/policy_test.go
rdrozhdzh: Thank you. I just copied this call from another test. Would you like me also to change to context.Background in other tests in this file?

huynhmb359:
It looks OK to me.

PR #82, opened 2017-10-18
Implement JSON AST parser

mstaneuka:
This is initial PR for JSON AST parser. It works, reduce memory more than 5 times and format configurable from PDP server (-pfmt option). 

There are several points that need to be improved.
1) Whereas policies and rules parsed as they come from the stream, some objects that assumed to be small are parsed as whole (algorithm, rule, obligation, condition etc).
2) Error messages format for jast looks not good: jcon helpers are used to process stream and it leads to incompatibility between jcon.boundError and ast.boundError (they have private bind methods)
3) jcon/helpers should be moved to separate package.

I'll proceed with these improvements.


vasili-v:
I think that there is possibility to remove boundError from common JSON parser library itself. So it will just propagate error returned from callbacks to upper level. While its own errors can be just of standard error type.

vasili-v:
Sooner or later we need to get rid of "small objects assumption" because some external PAP implementation can be not so good to keep small objects small.

mstaneuka:
All mentioned above points have been addressed

vasili-v:
Merging with decreased coverage as coverage enhance and some other grooming is going as separate PR

PR #94, opened 2017-11-22
Refactoring attrholder

UladzimirTrehubenka:


contrib/coredns/policy/attrholder_test.go
```diff
@@ -29,21 +29,21 @@ func TestActionFromResponse(t *testing.T) {
 		},
 		{
 			resp: &pdp.Response{Effect: pdp.Response_DENY, Obligation: []*pdp.Attribute{
-				{Id: "redirect_to", Value: "10.10.10.10"},
+				{AttrNameRedirectTo, "string", "10.10.10.10"},
```
rdrozhdzh: Id: AttrNameRedirectTo, Value: "10.10.10.10"

contrib/coredns/policy/dnstap/policytap.proto
```diff
@@ -1,10 +1,11 @@
-syntax = "proto2";
+syntax = "proto3";
```
rdrozhdzh: please copy the proto from 
https://github.com/Infoblox-CTO/golang-dnstap/blob/Coredns-Support/dnstap.pb/policytap.proto
to make sure they are the same

contrib/coredns/policy/attrholder_test.go
UladzimirTrehubenka: Fixed

contrib/coredns/policy/dnstap/policytap.proto
UladzimirTrehubenka: Fixed

PR #96, opened 2017-11-23
Move to single validation request per query

UladzimirTrehubenka:


johnbelamaric:
We need to clear this with @cricketliu. Allowing the query to be resolved before validation means we cannot prevent tunneling or other exfiltration, I believe.

UladzimirTrehubenka:
Exactly - so close PR

PR #99, opened 2017-12-05
PIP service rebase

yuewko:
**INCOMPLETE, FOR REVIEW ONLY**

Rebased PIP service changes.

Incomplete:
- connection manager need to be replaced with one that does discovery and load balancing


pip-service/connection.go
```diff
@@ -0,0 +1,56 @@
+package pip_service
```
vasili-v: The package as I commented above should be auto generated and contain only single gRPC wrapper file. It would be better to move the connection manager to PDP library.
Also the manager pretends to play two roles. It implements connection pool and it acts as a resolver for pip names. None of the roles have been completely implemented here. To make it production ready I'd move (at least for now) resolver role to DNS. So in pip selector URL `pip:<name>/<path>` the `<name>` would be just domain name of instance where PIP runs.
As connection pool the code should oversee connections and handle failed connections to prevent PDP from hanging.

pip-proto/service-gen.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
vasili-v: As mentioned earlier this should be replaced with `go:genereate` comments in related packages.

pdp/context.go
```diff
@@ -70,8 +73,8 @@ type Context struct {
 // name and value. If "f" function returns error NewContext stops iterations
 // and returns the same error. All pairs of attribute name and type should be
 // unique.
-func NewContext(c *LocalContentStorage, count int, f func(i int) (string, AttributeValue, error)) (*Context, error) {
-	ctx := &Context{a: make(map[string]interface{}, count), c: c}
+func NewContext(c *LocalContentStorage, p *ps.ConnectionManager, count int, f func(i int) (string, AttributeValue, error)) (*Context, error) {
```
vasili-v: I'd prefer to move connection management to policy storage. All references to pip reside in policy so we can add (similarly to `attributes` section) `pip` section to policy which will contain additional options for pip connections. The section will originate the manager (which I'd prefer to call pool for now).

pdp/ast/yast/target.go
```diff
@@ -27,6 +27,13 @@ func (ctx context) getAdjustedArguments(v interface{}, val pdp.Expression, attr
 
 		return a, attr, nil
 
+	case pdp.PIPSelector:
```
vasili-v: I'm going to remove ability to use selectors in target as set of targets from the same level policy sets, policies or rules is something which can be automatically indexed to speedup policy selection. The indexing isn't generally possible for expressions containing selectors (especially when network access is needed).

pdp/ast/yast/selector.go
```diff
@@ -23,42 +23,21 @@ func (ctx context) unmarshalSelector(v interface{}) (pdp.LocalSelector, boundErr
 		return pdp.LocalSelector{}, newSelectorURIError(s, ierr)
 	}
 
-	if strings.ToLower(ID.Scheme) == "local" {
-		loc := strings.Split(ID.Opaque, "/")
+	scheme := strings.ToLower(ID.Scheme)
+	loc := strings.Split(ID.Opaque, "/")
+
+	if scheme == "local" {
```
vasili-v: It's better use switch here or even better to have map of scheme to handler.

pdp/ast/yast/selector.go
```diff
@@ -23,42 +23,21 @@ func (ctx context) unmarshalSelector(v interface{}) (pdp.LocalSelector, boundErr
...
-		t, ok := pdp.TypeIDs[strings.ToLower(s)]
-		if !ok {
-			return pdp.LocalSelector{}, bindErrorf(newUnknownTypeError(s), "selector(%s.%s)", loc[0], loc[1])
-		}
-
-		if t == pdp.TypeUndefined {
-			return pdp.LocalSelector{}, bindErrorf(newInvalidTypeError(t), "selector(%s.%s)", loc[0], loc[1])
+		return makeLocalSelector(ctx, loc, m)
+	} else if scheme == "pip" {
+		if len(loc) != 2 {
```
vasili-v: I'd make validation for len(loc) > 0 here. The only PDP wants to know name of pip. Rest part of string can be treated differently by particular PIP implementation.

pdp/ast/jast/selector.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/infobloxopen/themis/pdp"
 )
 
-func (ctx context) unmarshalSelector(d *json.Decoder) (pdp.LocalSelector, error) {
+func (ctx context) unmarshalSelector(d *json.Decoder) (pdp.Expression, error) {
```
vasili-v: The same as for YAML parser - length of PIP url should not be generally limited and it's better make map string to selector parser instead of "if" statement.

pdp/ast/jast/selector.go
```diff
@@ -84,6 +85,22 @@ func (ctx context) unmarshalSelector(d *json.Decoder) (pdp.LocalSelector, error)
 		}
 
 		return pdp.MakeLocalSelector(loc[0], loc[1], path, t), nil
+	} else if strings.ToLower(id.Scheme) == "pip" {
+		loc := strings.Split(id.Opaque, "/")
+		if len(loc) != 2 {
+			return pdp.PIPSelector{}, newSelectorLocationError(id.Opaque, uri)
+		}
+
+		t, ok := pdp.TypeIDs[strings.ToLower(st)]
```
vasili-v: It looks differently comparing to YAML parser. Need to make similar.

pdp/ast/yast/parser_test.go
```diff
@@ -226,10 +226,9 @@ policies:
            - equal:
              - attr: strAttr
              - selector:
-                uri: "local:content/content-item"
+                uri: "pip:mcafee-ts/domain-category"
```
vasili-v: In the test check changed from local to pip. We need to check both of them. So please add pip selector but don't remove local selector.

yuewko:
`The server logic can be copied from local content and just serve JSON content using PIP protocol.`
Did you mean that the PIP server returns a JSON string instead of Attributes? If so, what's the reasoning behind that?

pdp/ast/jast/selector.go
yuewko: `length of PIP url should not be generally limited`

Did you mean this:
```
                loc := strings.Split(id.Opaque, "/")
                if len(loc) != 2 {
```
If that's what you mean...
The concept is that the url specifies `service_name/query_type`, and parameters passed in as attributes. I am not sure there's a need to allow a more general url.

pdp/ast/yast/parser_test.go
yuewko: Sorry. That was a hasty change and is totally unintentional.

pdp/local_selector.go
```diff
@@ -1,6 +1,8 @@
 package pdp
 
-import "fmt"
+import (
```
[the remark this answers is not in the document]
yuewko: Totally unintentional. Might have added other imports while coding.

pip-proto/service.proto
```diff
@@ -0,0 +1,32 @@
...
+
+package pip_service;
+
+service PIP {
+  rpc GetAttribute (Request) returns (Response) {}
+}
+
+message Attribute {
+  string id    = 1;
+  int32  type  = 2;
```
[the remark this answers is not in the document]
yuewko: Yes, I agree with consistency.

Actually, I'm surprised I didn't replicate the PDP definition for Attribute. Maybe I was questioning - why didn't we use `int32` in the first place?

pip-proto/service.proto
```diff
@@ -0,0 +1,32 @@
...
+  string value = 3;
+}
+
+message Request {
+  string query_type = 1;
```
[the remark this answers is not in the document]
yuewko: I agree that the PDP doesn't care about `query_type`.

However, in practice, the policy that uses the specific PIP selector is created specifically for that PIP selector AND that `query_type`. So, I am not sure it makes sense to specify `query_type` as an attribute.

vasili-v:
I'm talking about PIPCLI - an example of PIP implementation. Similarly Themis provides now PAPCLI - example of PAP implementation which just allows to upload given file to PDP server. The PIPCLI should provide the same functionality which local content provides but over network using PIP protocol of course getting and returning Attributes.

pdp/ast/jast/selector.go
vasili-v: Basically, we name this field URI and it brings following format: `scheme:[//[user[:password]@]host[:port]][/path][?query][#fragment]`. The first thing which matters to PDP is `scheme`. As soon as `scheme` is "local" meaning of rest is defined by PDP as PDP handles whole story. For "pip" scheme PDP should only care of `[user[:password]@]host[:port]]` to reach PIP server. Anything else it should pass to PIP server to not narrow possible implementations of latest.

pip-proto/service.proto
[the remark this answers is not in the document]
vasili-v: Now I got where query_type comes form. It should be URI instead holding value of URI field of corresponding selector.

PR #104, opened 2017-12-25
Added PIP service

UladzimirTrehubenka:


pdp-info/info.go
```diff
@@ -0,0 +1,156 @@
...
+
+	log "github.com/Sirupsen/logrus"
+)
+
+type streamObj struct {
+	stream PipService_GetRequestsStreamServer
+	err    chan error
+}
+
+type streamsPool struct {
```
james-xiang: Could you add some comments to explain details how streamsPool work ?
1. how many streamsPools are needed for multiple PIP Server Vendors ?  For example: we have McAfee PIP server, we have Symantec PIP server, etc. 
2. After a streamsPool is initialized, if a PIP server or all PIP servers are restarted, how to re-create the stream connections and put them into the streamsPool ?

UladzimirTrehubenka: We can use any count of PIP servers.
On PIP first connect PDP registers new stream with name that PIP returns.
If PIP is restarted - PDP removes broken stream and registers another one after PIP reconnect.

pdp-info/info.go
```diff
@@ -0,0 +1,156 @@
...
+func newStreamsPool() *streamsPool {
+	return &streamsPool{
+		streams: make(map[string][]streamObj),
+	}
+}
+
+func (p *streamsPool) Register(stream PipService_GetRequestsStreamServer) chan error {
+	cerr := make(chan error, 1)
+
+	req := &Request{Query: "register"}
```
james-xiang: it is difficult to understand the register flow here.  why don't ask the PIP to send the register request in the first command ? Because it is the PIP want to register itself to PDP.

UladzimirTrehubenka: Fixed

pdp-info/info.go
```diff
@@ -0,0 +1,150 @@
...
+		return cerr
+	}
+
+	inAttrs := resp.GetAttributes()
+	if len(inAttrs) != 1 {
+		cerr <- fmt.Errorf("Expected exactly one attribute, got: %d", len(inAttrs))
+		return cerr
+	}
+
+	name := inAttrs[0].GetValue()
```
james-xiang: it is not explicit here.  maybe let PIP send some string like "registration" to explicitly show the intention.

pdp-info/info.go
```diff
@@ -0,0 +1,150 @@
...
+	s := streamObj{stream: stream, err: cerr}
+
+	p.Lock()
+	defer p.Unlock()
+
+	if _, ok := p.streams[name]; !ok {
```
james-xiang: make sure the name is unique for each PIP instance

UladzimirTrehubenka: Name is unique for group of PIP. Map contains list of streams.

pdp-info/info.go
UladzimirTrehubenka: Do you mean to add verify GetId() that should be 'registration' (ATM is used id 'ident')?

pdp-info/info.go
```diff
@@ -0,0 +1,150 @@
...
+
+	list := p.streams[name]
+	if len(list) == 0 {
+		err = fmt.Errorf("PIP [%s] streams not registered", name)
+		log.Errorf(err.Error())
+		return
+	}
+
+	// streams random balancer
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
rdrozhdzh: Perhaps, random generator can be created once in streamsPool constructor

pdp-info/info.go
```diff
@@ -0,0 +1,150 @@
...
+	for {
+		stream, index, err := p.pool.GetStream(service)
+		if err != nil {
+			streamErr = err
+			break
+		}
+
+		req := &Request{Query: query, Attributes: attrs}
+		if err := stream.Send(req); err != nil {
+			p.pool.DeleteStream(service, index)
```
rdrozhdzh: at this point index may be invalid, since we unlocked the map mutex

pdp-info/info.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+
+		return resp.GetAttributes(), nil
+	}
+	return nil, streamErr
+}
+
+func (p *pipService) GetRequestsStream(stream PipService_GetRequestsStreamServer) error {
```
rdrozhdzh: imho, confusing name of the method

pdp/ast/yast/selector.go
```diff
@@ -7,59 +7,65 @@ import (
...
+	st, err := ctx.extractString(m, yastTagType, "type")
+	if err != nil {
+		return ret, bindErrorf(err, "selector(%s)", id.Opaque)
+	}
 
-		s, err := ctx.extractString(m, yastTagType, "type")
-		if err != nil {
-			return pdp.LocalSelector{}, bindErrorf(err, "selector(%s.%s)", loc[0], loc[1])
-		}
+	t, ok := pdp.TypeIDs[strings.ToLower(st)]
```
rdrozhdzh: would be nice to move common jast/yast logic into separate function to avoid copying of code

pdp-info/info.go
UladzimirTrehubenka: Good catch - I am going to use map instead list for fix this

PR #108, opened 2018-01-23
Made selector not target compatible

vasili-v:
Also target doesn't to swap expression arguments anymore


UladzimirTrehubenka:
Duplicates:
pdp/selector/registry.go
and
pdp/selector/init.go

PR #112, opened 2018-01-26
Attr assign backward compatible rebase

yuewko:
This PR include all the changes to add numerical functions in support of auto-scaling policies.

It also include changes that allow Attribute Assignment Expression to support both Immediate Values as well as Expressions


Makefile
```diff
@@ -154,6 +154,10 @@ build-egen: build-dir
 test-pdp: cover-out
 	$(AT)/pdp && $(GOTESTRACE)
 
+.PHONY: test-pdp-integration
```
vasili-v: I don't see a need of the integration tests it the form they have been added here. Themis package includes several different things - pdp, parsers, server, client libraries. What do the tests integrate? Why only some aspects of numbers, but not whole thing? I see those more like parser tests so please move them to related parser directory.

pdp/ast/jast/obligations.go
```diff
@@ -3,14 +3,15 @@ package jast
 import (
 	"encoding/json"
 
+	//	"fmt"
```
vasili-v: Don't comment it - just remove

pdp/ast/jast/obligations.go
```diff
@@ -24,11 +25,43 @@ func (ctx context) unmarshalObligationItem(d *json.Decoder) (pdp.AttributeAssign
...
+			if err != nil {
+				return bindError(err, k)
+			}
+		} else if delim == jparser.DelimArrayStart {
+			val, err := jparser.GetArray(d, "obligations")
+			e, err = ctx.unmarshalValueByTypeObject(a.GetType(), val)
+			if err != nil {
+				return err
+			}
+			/*
```
vasili-v: No need to comment code - just remove it. If you need it for later make other branch.

pdp/ast/jast/obligations.go
```diff
@@ -24,11 +25,43 @@ func (ctx context) unmarshalObligationItem(d *json.Decoder) (pdp.AttributeAssign
 			return newUnknownAttributeError(k)
 		}
 
-		e, err = ctx.unmarshalValueByType(a.GetType(), d)
+		delim, val, err := jparser.CheckObjectStartOrValue(d, "argument")
```
vasili-v: As I recall, we've agreed on no backward compatibility. So obligation is always an expression. It makes no need of CheckObjectStartOrValue function.

pdp/ast/jast/obligations.go
```diff
@@ -24,11 +25,43 @@ func (ctx context) unmarshalObligationItem(d *json.Decoder) (pdp.AttributeAssign
...
+			if err != nil {
+				return err
+			}
+		} else if delim == jparser.DelimObjectStart {
+			e, err = ctx.unmarshalExpression(d)
+			if err != nil {
+				return bindError(err, k)
+			}
+		} else if delim == jparser.DelimArrayStart {
+			val, err := jparser.GetArray(d, "obligations")
```
vasili-v: JAST is a stream parser. You need to work on token level. Don't try to get whole array and then parse it. Instead parse items as they appear with `CheckArrayStart` and by token iterations. However, if we follow no backward compatibility this isn't needed. Because obligation is always expression and you can just use existing expression parser here.

Makefile
yuewko: Maybe **integration** is a poor choice of word, but these are definitely not **unit** tests either. They don't just test the parsing, but the actual calculation too.

On your other point, I agree they definitely don't test *all* aspects of numbers. Please let me know what are the significant omissions and I'll definitely add those too.

pdp/ast/jast/obligations.go
yuewko: My bad. Will fix.

pdp/ast/jast/obligations.go
yuewko: My bad. It's no longer needed. I'll remove it.

pdp/ast/jast/obligations.go
yuewko: Yes, I know... I was very close to making it work when we discussed, so I dug deep. I thought it would be a good goal to avoid PAP changes.

My original plan was to have the numerical function changes (using expression only of obligations) as one PR, and the change to make it backward compatible a separate PR. This is so you can review to see whether the changes for backward compatibility is acceptable. (unfortunately, there was a lot conflicts when rebasing, so I bailed out...)

If you feel strongly that the backward compatibility changes should be pulled, I'll do that.

pdp/ast/jast/obligations.go
yuewko: Yes, totally understood. (please also see https://github.com/infobloxopen/themis/pull/112#discussion_r164430421)

pdp/ast/yast/obligations.go
```diff
@@ -23,9 +23,20 @@ func (ctx context) unmarshalObligationItem(v interface{}) (pdp.AttributeAssignme
 		return pdp.AttributeAssignmentExpression{}, newUnknownAttributeError(ID)
 	}
 
-	e, err := ctx.unmarshalValueByType(a.GetType(), v)
-	if err != nil {
-		return pdp.AttributeAssignmentExpression{}, bindError(err, ID)
+	var e pdp.Expression
+	m, err = ctx.validateMap(v, "obligation assignment")
+	if err == nil {
+		// if it's a map, assume it's an expression
```
vasili-v: Let's stick to no backward compatibility. It allows in addition to other simplifications to bypass some corner cases related to the assumption.

pdp/attribute.go
```diff
@@ -23,6 +23,10 @@ const (
 	TypeBoolean
 	// TypeString is string data type.
 	TypeString
+	// TypeInteger is integer data type.
```
vasili-v: Please add tests to attribute_test.go so we keep whole type converters tested.

pdp/ast/yast/obligations.go
yuewko: Sounds good - I'll take those code out.

BTW, I just remembered - one reason I tried to make it backward compatible is because the XACML specification stipulates using immediate value. I'm good if you don't feel that's critical.

Thanks!!

pdp/attribute.go
```diff
@@ -403,6 +463,12 @@ func (v AttributeValue) Serialize() (string, error) {
 	case TypeString:
 		return v.v.(string), nil
 
+	case TypeInteger:
+		return strconv.FormatInt(v.v.(int64), 10), nil
+
+	case TypeFloat:
+		return strconv.FormatFloat(v.v.(float64), byte('G'), 16, 64), nil
```
vasili-v: If no specific objections the -1 precision is better as it makes `FormatFloat` so many digits that `ParseFloat` returns exactly value `FormatFloat` gets. And you don't need to cast 'G' to byte. The changes make the line to look exactly as [above](https://github.com/infobloxopen/themis/pull/112/files#diff-f4fc264e7d4c0af26700a6f5fc900b41R278).

yuewko: Yes - my bad, again. I thought I changed them all to `-1`.
Will fix.

pdp/context.go
```diff
@@ -192,6 +202,41 @@ func (c *Context) calculateStringExpression(e Expression) (string, error) {
...
+func (c *Context) calculateFloatExpression(e Expression) (float64, error) {
+	v, err := e.Calculate(c)
+	if err != nil {
+		return 0, err
+	}
+
+	return v.float()
+}
+
+func (c *Context) calculateFloatOrIntegerExpression(e Expression) (float64, error) {
```
vasili-v: I'd like to see explicit conversion for each type. This assumes int-to-float implicit conversion. Having expression by argument type function selection on parsing stage such implicit conversion makes hard to follow function resolution rules. And we need to document it clearly as C++ does. I'd like to not mess with it.

pdp/target.go
```diff
@@ -192,7 +192,16 @@ type twoArgumentsFunctionType func(first, second Expression) Expression
 var TargetCompatibleExpressions = map[string]map[int]map[int]twoArgumentsFunctionType{
 	"equal": {
 		TypeString: {
-			TypeString: makeFunctionStringEqual}},
+			TypeString: makeFunctionStringEqual},
+		TypeInteger: {
+			TypeInteger: makeFunctionIntegerEqual},
+		TypeFloat: {
```
vasili-v: Actually, here you missed float+int and int+float -> makeFunctionFloatEqual cases if you want to follow implicit conversion. However, I don't think it's worth to make it.

Makefile
vasili-v: I prefer to test calculations inside pdp constructing policies with Make* functions. Basically, we don't need much integration inside single repository.

yuewko: OK. Let me give it a shot.

pdp/target.go
yuewko: Sounds good!

Actually, I thought about that. The options are:
- Promote the int to float and then compare as floats. This typically is not reliable. It might makes sense for greater than, but not for equal, IMHO.
- Cast/Round the float to int and then compare as int. This is potentially very confusing for the users.

pdp/context.go
yuewko: I understand your concern.

I just wanted to point out that `calculateFloatOrIntegerExpression` is specifically designed for the float numerical functions, where I specifically want to promote int arguments to float. Without it, the code will be duplicated for each of the arguments for each of the float functions.

Maybe I can move/rename that function if you feel it doesn't below to that group of `calculate` functions.

Please let me know how you'd like to proceed.

yuewko:
https://github.com/infobloxopen/themis/pull/112/commits/02aa45c562e095884f4196f81159567ed72e149d addressed the following review comments:

- https://github.com/infobloxopen/themis/pull/112#pullrequestreview-92190947
- https://github.com/infobloxopen/themis/pull/112#pullrequestreview-92191337
- https://github.com/infobloxopen/themis/pull/112#pullrequestreview-92203254
- https://github.com/infobloxopen/themis/pull/112#pullrequestreview-92207587

pdp/ast/jast/obligations.go
yuewko: Done

pdp/ast/jast/obligations.go
yuewko: Done

pdp/attribute.go
yuewko: Done

pdp/attribute.go
yuewko: Done

yuewko:
Added unit tests for numerical functions to address https://github.com/infobloxopen/themis/pull/112#pullrequestreview-92190461

yuewko:
@vasili-v , I believe I have addressed all the review comments, with the exception of documentation updates (i.e. README and Examples).

Please review when you have a chance.

yuewko:
@vasili-v , the latest commit includes README updates and additional examples. This should have addressed all your comments.

Thanks!!

yuewko:
OK. Thanks for looking at it.

PR #113, opened 2018-01-29
Added 'ident' and 'passthrough' options for policy plugin

UladzimirTrehubenka:


contrib/coredns/policy/policy.go
```diff
@@ -140,6 +142,12 @@ func (p *policyPlugin) parseOption(c *caddy.Controller) error {
 
 	case "transfer":
 		return p.parseTransfer(c)
+
```
vasili-v: I think it's time to update README with all new options.

contrib/coredns/policy/policy.go
```diff
@@ -466,8 +481,14 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
 		sendExtra bool
 	)
 
-	debugQuery := p.decodeDebugMsg(r)
 	qName, qType := getNameType(r)
+	for _, s := range p.passthrough {
+		if strings.HasSuffix(qName, s) {
+			return plugin.NextOrFailure(p.Name(), p.next, ctx, w, r)
```
rdrozhdzh: Hm, what should we report with DNStap this case?

UladzimirTrehubenka: I think no

contrib/coredns/policy/policy.go
UladzimirTrehubenka: Updated

contrib/coredns/policy/README.md
```diff
@@ -25,18 +33,30 @@ Param START and END (last data byte index + 1) allow to get separate part of edn
 
 Option debug_query_suffix SUFFIX (should have dot at the end) enables debug query feature.
 
+Option streams set gRPC streams count for PDP connection.
+
+Option transfer defines set of attributes (from domain validation response) that should be inserted into IP validation request.
+
+Option ident set string that is used for debug query response as unique id for determine what CoreDNS instance replies on the request.
```
vasili-v: Maybe `debug-id`?

vasili-v:
👍

PR #121, opened 2018-02-14
Policy Plugin: delay closing connection to PDP

rdrozhdzh:


rdrozhdzh:
The problem is that coredns continue validating DNS queries (queries in progress) at the moment when restart event is fired. In restart handler we close connection to PDP and coredns returns SERVFAIL for the remaining queries because of lost connection to PDP.
My fix allows the remaining queries to complete successfully

vasili-v:
The fix still doesn't work for all cases. Even worse it makes an issue harder to catch while testing and debugging leaving chance to get it in production. Additionally, resources allocated for connection are held for 30 seconds more after each restart. Having ability to get 1 restart per 5 seconds we can have 6 copies of pdp client alive.

PR #124, opened 2018-02-19
starting to use a separate flag set for parsing CLI options for PDP

skudriavtsev:
... to avoid flag conflicts with other packages that we are importing ('v' flag in glog used by k8s as an example).


vasili-v:
Please point to source introduced the conflict.

vasili-v:
Closed as the conflict isn't a Themis issue.

PR #123, opened 2018-02-15
Fixed: No information in logs about category when its added to CF

mingkaic:
Change to make logs more informative:
Format from 
```
DEBU[0006]` Policy update                                 update=policy update: 93012340-27cc-47af-97a6-59cd654cf2cc - 7b9096d1-3ad1-4fbf-940d-193d4df780e1
commands:
- Add ("Default"/"query"/"DFPRNetworks"/"0000002e"/"Default")
- Add ("Default"/"query"/"DFPRNetworks"/"CIDRNetworks"/"0000002e"/"Default")
```
to
```
DEBU[0006] Policy update                                 update=policy update: 94db536a-f402-4ea9-a9cb-a3e3445bddf9 - 3664a605-97d3-427d-ac21-4acf8de56f96
commands:
- Add policy "Default" rules([Default News Blogs Tobacco Drugs]) to
  path: ("Default"/"query"/"DFPRNetworks"/"00000001"/"Default")
- Add policy "Default" rules([Default News Blogs Tobacco Drugs]) to
  path: ("Default"/"query"/"DFPRNetworks"/"CIDRNetworks"/"00000001"/"Default")
```

PR #125, opened 2018-02-20
Make PDP Logs Add CF More Descriptive 

mingkaic:
Continuation of https://github.com/infobloxopen/themis/pull/123

### DIFF
--- a/pdp/policy.go
+++ b/pdp/policy.go
@@ -59,8 +59,16 @@ func NewPolicy(ID string, hidden bool, target Target, rules []*Rule, makeRCA Rul
 }
 
 func (p *Policy) describe() string {
+	ruleIDs := make([]string, len(p.rules))
+	ruleIdx := 0
+	for _, rule := range p.rules {
+		if ruleID, ok := rule.GetID(); ok {
+			ruleIDs[ruleIdx] = ruleID
+			ruleIdx++
+		}
+	}
 	if pid, ok := p.GetID(); ok {
-		return fmt.Sprintf("policy %q", pid)
+		return fmt.Sprintf("policy %q rules(%v)", pid, ruleIDs[:ruleIdx])
 	}
 
 	return "hidden policy"
--- a/pdp/policy.go
+++ b/pdp/policy.go
@@ -1,6 +1,12 @@
 package pdp
 
-import "fmt"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+const nDisplayRuleLimit = 5
 
 // RuleCombiningAlg represent abstract rule combining algorithm. The algorithm
 // defines how to evaluate policy rules and how to get paticular result.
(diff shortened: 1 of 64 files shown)


mingkaic:
@vasili-v Regarding the control interface extension, I'm not sure whether I should include this feature to the upcoming debug enhancement involving our internal codebase or implement a control interface as part of this fix.

pdp/storage.go
```diff
@@ -143,6 +143,14 @@ func (c *command) describe() string {
 		}
 	}
 
+	if nil != c.entity {
```
vasili-v: Golang doesn't allow assignment in condition so you can do it in natural way `c.entity != nil`.

pdp/storage.go
```diff
@@ -143,15 +143,17 @@ func (c *command) describe() string {
 		}
 	}
 
+	opPath := strings.Join(qpath, "/")
 	if nil != c.entity {
 		if evaluable, ok := c.entity.(Evaluable); ok {
-			if id, ok := evaluable.GetID(); ok {
-				qpath = append(qpath, strconv.Quote(id))
-			}
+			// change to something on path
```
vasili-v: This and one below are both misleading comments. Please remove or rewrite them. We don't change path or something on path. We always change policies we have. The difference is that some updates contain entity we want to bring (as for "add") and for other we use entity which we already have (as for "del").
Additionally, what is more important, it makes your messages asymmetrical. It was symmetrical before because both didn't expose item in question only its location. Now it should as well say something like that:
"Adding/Deleting policy_set/policy/rule with following policies/rules at ...". I understand that getting information for delete is challenging for delete operation but still think that it's more natural to make them symmetrical.

pdp/storage.go
```diff
@@ -143,15 +143,17 @@ func (c *command) describe() string {
 		}
 	}
 
+	opPath := strings.Join(qpath, "/")
 	if nil != c.entity {
 		if evaluable, ok := c.entity.(Evaluable); ok {
```
vasili-v: Why we don't case for rules here? Evaluable is an interface only policy and policy set while for rule message remains as before. We could say for example that we are adding permit or deny rule.

pdp/policy.go
```diff
@@ -63,12 +69,15 @@ func (p *Policy) describe() string {
 	ruleIdx := 0
```
vasili-v: Policy set also has describe method. Why didn't change it as well?

vasili-v: Line above allocates array of the same length as array of rules has. It can be to big array of for example million items. You should allocate no more than nDisplayRuleLimit + 1.

pdp/policy.go
```diff
@@ -63,12 +69,15 @@ func (p *Policy) describe() string {
...
 	for _, rule := range p.rules {
```
vasili-v: You don't need to iterate over all rules. Find first nDisplayRuleLimit - 1 rules and then from the end find another one. Also you need to process case when number of rules less than nDisplayRuleLimit separately.

mingkaic:
It seems like this description would be a lot of work if the rules are filled with hidden rules.
For now, is it ok just to merge it? I'm adding a log detailed command in the Infoblox-CTO repo, then adding it here in a later pull request. Does this sound good?

vasili-v:
Don't think so it's better to make all the changes regarding to the task right here. I'm also expecting changes to storage.go and policy_set.go according to my comments.

mingkaic:
@vasili-v I realized that `func (r Rule) GetID() (string, bool)` in rule.go actually returns the id string even if the rule is hidden. Would it be problematic/break security, if we printed hidden id when debugging? If we can just print it regardless it would simplify the logic significantly

mingkaic:
This bug will be resolved in PR https://github.com/infobloxopen/themis/pull/127

PR #127, opened 2018-02-21
Make Pdpserver Use LoggerField for Extended Log Control

mingkaic:


pdpserver/server/service.go
```diff
@@ -118,9 +118,7 @@ func (s *Server) rawValidate(p *pdp.PolicyStorage, c *pdp.LocalContentStorage, i
 		return pb.Response_INDETERMINATE, []error{err}, nil
 	}
 
-	if s.opts.logger.Level >= log.DebugLevel {
```
mingkaic: I don't understand the point of this check. Isn't Debug the highest verbosity level anyways?
Also logrus should now support this check. Is this for a previous bug on logrus' side?

mingkaic:
This pull accomplishes the following:
1. Some log calls do not use server logger. This does not guarantee the same logger instance between private and Themis repo
2. Make server logger an interface to support potential Logger upgrades in private repo

pdpserver/server/service.go
vasili-v: Please revert the change. It's an optimization. WithFields function takes insane amount of time and with no the check it's called each time even if log is higher than DebugLevel. Logrus checks log level on later stage inside Debug function which is (as I said) called after WithFields.

mingkaic: Is the time issue on our side or on logrus' side? If it is on our side because describing the context is a lot of work, perhaps it's a good target for the upcoming detailed control feature.

UladzimirTrehubenka: See https://github.com/sirupsen/logrus/issues/125

UladzimirTrehubenka:
I don't follow were is abstraction here?
You replace
log.WithField("update", u).Debug("Policy update")
with
s.opts.logger.WithField("update", u).Debug("Policy update")
But this means that s.opts.logger should use logrus API convention and we need to wrap this call in case of using logger differ

mingkaic:
@UladzimirTrehubenka I'm extending the logger on Infoblox-CTO repo with a detailed mode. On themis-side, the detailed toggle can show more or less of the log as mentioned by Vasili in https://github.com/infobloxopen/themis/pull/123. This is necessary because some describe functions seem to be more verbose at the cost of time. A detailed mode allows proper debugging without sacrificing time.
Should I commit the changes in CTO before this one?

mingkaic:
Vasili called this "extending control interface" but the debug endpoint is more appropriate because it deals mainly with log control.

mingkaic:
A use case for this is when you want Info mode to print out just the policy ID, but when a bug arise, you want to check the rules under that policy. With only logrus.Logger, you can print out twice (one in Info with just the policy ID, and one in Debug mode with policy ID and a certain number of rules). If you can specify in the extended logger how many rules to show.
In the logrus use case, the code is more bloated and prone to error and less flexible. In the extended version, you only print once and can specify how detailed.

pdpserver/server/log_mgr.go
```diff
@@ -0,0 +1,19 @@
+package server
```
vasili-v: Please rename log_mgr.go to logger.go. There is no management just additional logger interface.

pdpserver/server/log_mgr.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/Sirupsen/logrus"
+)
+
+type LevelCheckable interface {
+	GetLevel() logrus.Level
+}
+
+func GetLogLevel(logger logrus.FieldLogger) logrus.Level {
+	if lgr, ok := logger.(*logrus.Logger); ok {
```
vasili-v: Everywhere when you need to distinct more than one type it's simpler to use type switch. 
```go
switch log := logger.(type) {
case *logrus.Logger:
	return log.Level

case LevelCheckable:
	return log.GetLevel()
}
```

pdpserver/server/log_mgr.go
```diff
@@ -0,0 +1,19 @@
...
+		return lgr.Level
+	} else if lgr, ok := logger.(LevelCheckable); ok {
+		return lgr.GetLevel()
+	}
+	// else
```
vasili-v: It's quite obvious that here we have default type so don't need to put the comment.

pdpserver/server/log_mgr.go
```diff
@@ -0,0 +1,19 @@
...
+	return 0 // return the lowest verbosity to be safe
```
vasili-v: As for me it's better to use predefined constant instead of plain number. I'd put directly `logrus.PanicLevel`. But it even more readable to define constant here.

```go
const lowestVerbosityLevel = logrus.PanicLevel

func GetLogLevel(logger logrus.FieldLogger) logrus.Level {
	...

	return lowestVerbosityLevel
}

File: pdpserver/server/log_mgr.go [hunk 6]
```diff
@@ -0,0 +1,19 @@
+package server
+
+import (
+	"github.com/Sirupsen/logrus"
+)
+
+type LevelCheckable interface {
```
vasili-v: All exported symbols which names start with capital letter should have documentation comments. You can run `golint` or `golint -min_confidence 0` to see if your code has that kind of issues.

vasili-v: Additionally, it would be good to mention this feature somewhere in README. But for now we seem haven't created section in README for embedding pdpserver as a library.

mingkaic:
For the sake of time, I think it's best to add a section for pdpserver as a library as part of the update for https://github.com/infobloxopen/themis/pull/125. You can review that and the logrus enhancement in Infoblox-CTO on Tuesday/Wednesday. This pull request should be merged by Monday/Tuesday. 

I'm still new at this, so thank you for all the help.

mingkaic:
@vasili-v can you code review this? Thanks

pdpserver/server/logger.go
```diff
@@ -0,0 +1,40 @@
...
+//	  that the logger does not consume through GetDetail
+type DetailedInfo interface {
+	GetDetail(nShow uint) string
+	FilterLevel() logrus.Level
+	String() string
+}
+
+// compile-time interface check
+var (
+	_ DetailedInfo = pdp.PolicyUpdateDetail{}
```
vasili-v: The line gives a clue that DetailedInfo interface belongs to pdp not to here. But is it needed at all? I see it as a way to extend number of log levels to different dimension. Why does it appear that levels which logrus provide out of the box aren't enough? I still prefer to not solve the issue by hacks as this. But instead introduce proper set debug tools. One of them should be able to get policy by id from pdp.

PR #138, opened 2018-03-01
add policy/rule query by id for debugging enhancement

mingkaic:
Additional changes:
- Reverted expensive and verbose logging on policy changes in https://github.com/infobloxopen/themis/pull/123
- Updated README to match changes


mingkaic:
@vasili-v can you review this? Can you verify whether this method of debug control is useful?

markinos:
As the main programmer on authorization, I work with the PDP a lot. I find these commits and new method of debugging to be potentially very useful.

README.md
```diff
@@ -891,16 +891,32 @@ INFO[2190] Got new data stream
...
+- Add Path ("domain-addresses"/"bad"/"example.com")
 INFO[2190] Got apply command
 INFO[2190] Content update has been applied               cid=content curr-tag=93a17ce2-788d-476f-bd11-a5580a2f35f3 id=5 
 ...
 ```
 
-Contents with different ids and policies can be updated independently and in paralel.
+Contents with different ids and policies can be updated independently and in parallel.
+
+## Debugging Options
```
vasili-v: I think that as policies has tree-like structure it's better to reflect the fact in API.

So `GET /storage/rootPID/innerPID/moreInnerPID/...` gives exact policy or rule by path (and you don't need to provide separate endpoints for policies, policy sets and rules). You can control how many nested policies and rules you get with query parameter. for example `GET /storage/rootPID?depth=2` gives root policy itself and all its immediate children.

For search I suggest to have separate endpoint `/search` with query parameter id. It should use "contains" comparison to get matching policies. And return list of full paths to them. So `GET /search?id=xxx` returns "/yxxxz", "/yxxxz/xxx" but not "/yxxxz/yyy". In unix shell command `find / -name "*xxx*"` works very similar to that. Here you also can add parameters such as depth to limit nesting search goes down and limit parameter to limit number of results.

You need to take special measures to not eat all memory during rendering policies as a response.

pdpserver/server/query.go
```diff
@@ -0,0 +1,166 @@
...
+					}
+				}
+				fmt.Fprint(w, strings.Join(policyIDs[:pubIdx], ", "))
+			}
+		})
+	var err error
+	go func() {
+		err = http.ListenAndServe(s.opts.queryEP, storageQueryRouter)
+	}()
+	if err := checkEP(s.opts.queryEP, func() error { return err }); err != nil {
```
vasili-v: What do you want to achieve by this?

mingkaic:
New Overview for the changes:
- Added Iterable interface for PolicySet, Policy, and Rule to traverse the tree
- Query the Tree using DFS to avoid memory overhead.
- API reflects the tree structure through path parameters

mingkaic:
@vasili-v can you review the changes?

pdp/storage.go
```diff
@@ -57,6 +57,101 @@ func (s *PolicyStorage) NewTransaction(tag *uuid.UUID) (*PolicyStorageTransactio
...
+				return fmt.Sprintf("{%s,\"elems\":[%s]}", idInfo,
+					strings.Join(children[:cIdx], ","))
+			}
+		} else if iter.NextSize() > 0 {
+			return fmt.Sprintf("{%s,\"elems\":\"...\"}", idInfo)
+		}
+		return fmt.Sprintf("{%s}", idInfo)
+	}
+	return ""
+}
```
mingkaic: I know GetSubtree returning JSON string is weird, but it seems like the cheapest and simplest way to limit the height of the tree for our purpose. The recursive depth is min(depth, height of the tree). An alternative way is to create temporary iterables (as to not change the storage tree), then serialize it in our query API on pdpserver. This is more generalized and reusable later.

pdpserver/server/query.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/infobloxopen/themis/pdp"
+	"github.com/julienschmidt/httprouter"
+)
+
+const (
+	storageQueryCommands = `PDP Query API
+
+GET /root
+GET /storage/<path...>?depth=<depth> [depth is any value of range 0 .. n. defaults to 0]
+GET /find/<id>/<start path...> [ensure to end with "/" if start path is root]
```
vasili-v: Still not convinced with design. Natural way to make find query is:
```
GET /find/<path>?id=<id>
```

pdpserver/server/query.go
```diff
@@ -0,0 +1,131 @@
...
+	maximumRetries  = 100
+	maxUint32       = 2147483647
+	defaultDepthMsg = "proceeding with depth=0"
+)
+
+func queryIndexHandler(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	fmt.Fprint(w, storageQueryCommands)
+}
+
+func (s *Server) listenQuery() error {
```
vasili-v: The function name is misleading and too generic. Too generic because doesn't specify what query it listen for. Should be something like `listenPolicyStorage`. Misleading because it's ListenAndServe. Which it shouldn't be. We currently have a couple of examples how to do it (health check and profile server). Basically, maybe it's good point to gather them all together at single http port.

pdp/context.go
```diff
@@ -317,3 +317,20 @@ type Evaluable interface {
 	setOrder(ord int)
 	describe() string
 }
+
+// Iterable interface defines all PDP entities having some parent-child relationship
```
vasili-v: I don't think that `Iterable` interface is good idea. It's simpler just to add Marshal to Evaluable, Rule, Expression and other involved structures.

pdp/storage.go
vasili-v: That's why iterable isn't good. Recursive marshaling can easily limit depth by passing current depth to each marshal function.

vasili-v: Anyway it should be rewritten to not return string of arbitrary size. Instead marshallers  should be able to use fixed size buffer and send the buffer to output as soon as it's full. It can be done with callback or channel approach. Latest is more idiomatic while the callback seems to be more performant.

pdpserver/server/query.go
```diff
@@ -0,0 +1,131 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/infobloxopen/themis/pdp"
+	"github.com/julienschmidt/httprouter"
```
vasili-v: I don't think we need the router as our case is extremely simple.

vasili-v:
Closing this as we started over with step by step implementation.

mingkaic:
@vasili-v Regarding your comments in https://github.com/infobloxopen/themis/pull/160, isn't the implementation we're looking for basically this iterable interface renamed as marshal interface with baked in Marshal function? This would satisfy query and efficient dumping functionality?

PR #139, opened 2018-03-05
 Move to use fastrpc instead grpc for PDP service [for debug only]

UladzimirTrehubenka:


vasili-v:
Closing as not actual for now.

PR #140, opened 2018-03-06
Check for pass through only if it is not a debug query

huynhmb359:
Fix JANUS-13598 by adding a check for pass-through only if the query is not a debug query.


UladzimirTrehubenka:
See https://github.com/infobloxopen/themis/commit/137545bdc7e5287049e2d4230a932ade28400ce2

PR #144, opened 2018-03-13
Improved passthrough logic for policy plugin

UladzimirTrehubenka:


rdrozhdzh:
also would be nice to have a UT for pass-through

PR #146, opened 2018-03-15
Move flag.Parse() to main()

yuewko:
Move flag.Parse() to main() so that other packages can use init()
to create flags just like config.go (in this case, ngp-policy-engine/pip/plugin/mcafee)

### DIFF
--- a/pdpserver/config.go
+++ b/pdpserver/config.go
@@ -23,7 +23,9 @@ var policyParsers = map[string]ast.Parser{
 }
 
 type config struct {
+	logVerbosity        int
 	policy              string
+	policyFmt           string
 	policyParser        ast.Parser
 	content             stringSet
 	serviceEP           string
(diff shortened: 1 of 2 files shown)


vasili-v:
Closing the pull request as PDP sever executable implementation is just an example and shouldn't meet any requirements which third party closed source projects can demand.
Moreover Themis provides server as a library so there are many ways to test plugins with it without running as side process.

johnbelamaric:
If we want it done as a library, it shouldn't be using `flag` directly, because that forces the developer that is importing it to use all those flags. We could have a separate method that they can call to setup the flags if they want to. But the library initialization itself should be through either a Config struct or through Options (or a combination).

yuewko:
Correct me if I'm wrong, I THINK @vasili-v is trying to say the pdpserver is just an example, which is the piece that has the flag code.

I think @vasili-v is referring to the pdp package as the library, which is the Themis library that others would be using. The pdp package does not depend on the config code.

Having said that, as I mentioned to @vasili-v in a separate thread - Ialthough pdpserver is just an example, I think we should try to adopt best practice whenever possible.

And, I thing what you suggested is best practice.

PR #148, opened 2018-03-16
Coredns-plugin: Get User SourceIP when resolving through DNS-Dist

mingkaic:
I'm making this change to solve issues dealing with the following use case:

- Use 1: User (addr X) calls Coredns directly to resolve their query. In this scenario, the SourceIP takes address X. (Expected behavior)
- Use 2: User (addr X) calls Coredns through DNS-Dist (addr Y). In this scenario, the SourceIP takes address Y. (Unexpected behavior)

We want both use cases to have SourceIP take X (user IP).
Since the source IP is lost along the way to DNS-Dist, Thong proposed passing the user's IP as an edns0 option with code 0xfff5.
These changes allows Coredns-plugin to read this option.

### DIFF
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -44,6 +44,8 @@ const (
 	actCount
 )
 
+const srcIPedns0code = 65525 // edns0 encoding of 0xfff5
+
 var actionConv [actCount]string
 
 func init() {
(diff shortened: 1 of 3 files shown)


contrib/coredns/policy/policy.go
```diff
@@ -44,6 +44,8 @@ const (
 	actCount
 )
 
+const srcIPedns0code = 65525 // edns0 encoding of 0xfff5
```
vasili-v: Why don't:
```
const srcIPedns0code = 0xfff5
```
?

UladzimirTrehubenka:
See https://github.com/infobloxopen/themis/pull/149

PR #153, opened 2018-03-26
Made LocalAddr and RemoteAddr to return addresses of real writer

vasili-v:
Method `RemoteAddr` of `dns.ResponseWriter` is used by forward plugin to decide which protocol UDP or TCP to use with upstream server. This convention isn't easy to find so I also suggest to make other methods of policy plugins' writer to panic instead of silently return:
```go
func (w *writer) Close() error                  { panic(errors.New("not implemented") }
func (w *writer) TsigStatus() error             { panic(errors.New("not implemented") }
func (w *writer) TsigTimersOnly(b bool)         { panic(errors.New("not implemented") }
func (w *writer) Hijack()                       { panic(errors.New("not implemented") }
func (w *writer) LocalAddr() (la net.Addr)      { return w.w.LocalAddr() }
func (w *writer) RemoteAddr() (ra net.Addr)     { return w.w.RemoteAddr() }
func (w *writer) WriteMsg(m *dns.Msg) error     { w.Msg = m; return nil }
func (w *writer) Write(buf []byte) (int, error) { panic(errors.New("not implemented") }
```

### DIFF
--- a/contrib/coredns/policy/policy.go
+++ b/contrib/coredns/policy/policy.go
@@ -607,7 +607,7 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
 	qName, qType := getNameAndType(r)
 	for _, s := range p.passthrough {
 		if strings.HasSuffix(qName, s) {
-			responseWriter := new(writer)
+			responseWriter := newWriter(w)
 			_, err = plugin.NextOrFailure(p.Name(), p.next, ctx, responseWriter, r)
 			r = responseWriter.Msg
 			status = r.Rcode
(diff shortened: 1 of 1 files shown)


contrib/coredns/policy/policy.go
```diff
@@ -808,14 +808,21 @@ func extractRespIP(m *dns.Msg) (address string) {
...
+	}
+}
+
 func (w *writer) Close() error                  { return nil }
 func (w *writer) TsigStatus() error             { return nil }
 func (w *writer) TsigTimersOnly(b bool)         { return }
 func (w *writer) Hijack()                       { return }
-func (w *writer) LocalAddr() (la net.Addr)      { return }
-func (w *writer) RemoteAddr() (ra net.Addr)     { return }
+func (w *writer) LocalAddr() (la net.Addr)      { return w.w.LocalAddr() }
```
nsinghisking: what is the impact of this change

vasili-v:
Closing as more proper way to use writer from [coredns/plugin/pkg/nonwriter](https://github.com/coredns/coredns/tree/master/plugin/pkg/nonwriter).

PR #160, opened 2018-04-12
Add Marshal Interface and Implement on Attribute

mingkaic:


vasili-v:
I get idea of the changes but the implementation has some questionable points:
1. User can go down only to policy set, policy or rule. So `Attribute` shouldn't satisfy marshaller interface. It can have private marshal method however which can only be called by its owners inside pdp package. Acctually, attribute works in PDP in the way you don't need to marshal it at all.
2. You don't need Size method. All strings and bytes sequences already have size as part of data type so writer can work with it.
3. Your `CappedBuffer` is new but not so feature full `bufio.Writer`. 

Basically, what I expect is making Marshaller interface with only method `Marshal(w io.Writer, depth int) error`. We don't even need to know if writer is buffered or not. We only need to call w.Write as soon as it make sense. For example, you shouldn't write some policy as single JSON object to writer. Instead you just need to write all its simple fields and then iterate through children with Marshal(w, depth-1) until depth > 0.

Now you need to start over with the Marshaller I mentioned and for example PolicySet struct. Implement Marshal method which only converts fields one by one and writes it with writer method. Skip child marshaling for now and create another PR.

I close this as per description above.

PR #161, opened 2018-04-12
Added custom flags type

vasili-v:


vasili-v:
This is still preview PR. Need to add flags value YAST/JAST parsing and update README documentation. When is ready and committed essentially outdates Themis wiki in part of adding new built-in types. Later I'm going to create another example of adding "complex" type. As we already have added "integer".

vasili-v:
The PR is complete now. Themis wiki in part of adding new built-in types still needs an update but it isn't in the scope of the change.

README.md
```diff
@@ -38,13 +38,40 @@ Some application may require more details for particular decision. For example i
...
+**Boolean** value is accepted as "1", "t", "T", "TRUE", "true", "True", "0", "f", "F", "FALSE", "false", "False" and serialized to "true" and "false". **Integer** value is a decimal number in range [-9223372036854775808, 9223372036854775807]. **Float** value can be specified using decimal format (e.g. 3.1416) or scientific notation (e.g. 6.022E+23). **Address** accepted in dotted decimal ("192.0.2.1") form or in IPv6 ("2001:db8::68") form and serialized respectively. **Network** is accepted as a CIDR notation IP address and prefix (for example "192.0.2.0/24" or "2001:db8::/32"). **Domain** name is accepted as string of labels separated by dots which satisfies to RFC1035, 2181 and 4343 requirements. **Set of strings**, **set of domains**, **set of networks** and **list of strings** aren't accepted in request context but can appear in response's obligations as comma separated list of values.
+
+User can define her custom type based on **flags** metatype. A value of the type can be any combination of listed flags. PDP allows to define up to 64 flags for a type. Values can't appear in request or returned as obligations.
```
yuewko: `Values can't appear in request or returned as obligations.`

I believe the above sentence is NOT specifically talking about **flags**. Rather, it is talking about **obligations**. If so, I believe it should be moved to couple of paragraphs earlier where it talk about **obligations**.

vasili-v: I talk here about **flags**. It's meta type and particular instance needed to be communicated to client beforehand to make client be able to send a value of the type to server or get obligation with such value back. As we don't have "type handshake" client can't send flags to server as part of request. And server can't send flags to client as an obligation. So it isn't just about obligations.

yuewko: @vasili-v , thanks for the explanation! I actually played with it a bit - which helped me understand your points. IMHO, it might help if:
- say `*Flag* values can't appear....`
- There is an example below that showed how **flags** can be used in an obligation. Maybe you can refer to that example here, and point out how it can be used in obligations.

vasili-v:
Closing as the code integrated with [PR #169](https://github.com/infobloxopen/themis/pull/169).

PR #162, opened 2018-04-13
Implement Marshal for Evaluables + Rule

mingkaic:


pdp/marshal.go
```diff
@@ -0,0 +1,9 @@
+package pdp
+
+import "io"
+
+// StorageMarshal interface defines functions
+// to capturing storage state information
+type StorageMarshal interface {
+	DepthMarshal(out io.Writer, depth int) error
```
vasili-v: So what about naming?

pdp/policy.go
```diff
@@ -27,8 +30,12 @@ var (
 	// of the algorithm. Contains only algorithms which require parameters.
 	RuleCombiningParamAlgs = map[string]RuleCombiningAlgMaker{
 		"mapper": makeMapperRCA}
+
+	errHiddenPolicy = fmt.Errorf("Attempting to marshal hidden policy")
```
vasili-v: While we have only several marshaling functions an error can appear at any nesting level. It is good to log where exactly it goes from. PDP has specific approach implemented with boundError struct (see error.go file). To use it you need to define your specific error with error.yaml (there are a lot of examples here so I hope you can get in easily). `go generate` to update errors.go file after changes in errors.yaml. Then you need to make error with generated new&lt;WhateverYouPutHere&gt; function and instead of `return err` put `return bindError(err, "whatever context")` or `return bindErrorf(err, "what%s context", "ever")`.

pdp/policy.go
```diff
@@ -243,6 +250,42 @@ func (p *Policy) delChild(ID string) (*Policy, error) {
 	return p.updatedCopy(rules, algorithm), nil
 }
 
+// DepthMarshal implements StorageMarshal
+func (p Policy) DepthMarshal(out io.Writer, depth int) error {
+	if depth < 0 {
+		return fmt.Errorf("depth must be >= 0, got %d", depth)
+	}
+	if p.hidden {
+		return errHiddenPolicy
```
vasili-v: Don't think we want to fail on hidden policy. It's better to show it or silently skip. To choose behavior we can add special request option like `/storage/x/y/z?show=all` later. For now it's simpler to not skip hidden policies.

pdp/policy.go
```diff
@@ -243,6 +250,42 @@ func (p *Policy) delChild(ID string) (*Policy, error) {
...
+	}
+	_, err := out.Write([]byte(fmt.Sprintf(policyFmt, p.ord, p.id)))
```
vasili-v: Please double check that `Write` isn't allowed to write less than passed without any error. If it's allowed you need to make sure that all bytes have been written before going further.

Also `p.id` can be any string and contain JSON special characters as `"`, `,` and so on. So it's better to use json.Marshal here to format fields as `p.ord` and `p.id` and than put it them with `{"ord": %s, "id": %s, "rules": [` format.

pdp/policy.go
```diff
@@ -243,6 +250,42 @@ func (p *Policy) delChild(ID string) (*Policy, error) {
...
+				if err = r.DepthMarshal(out, depth-1); err != nil {
+					return err
+				}
+				firstRule = i
+				break
+			}
+		}
+		for _, r := range p.rules[firstRule+1:] {
+			if _, ok := r.GetID(); ok {
+				out.Write([]byte{','})
```
vasili-v: This also can fail.

pdp/policy_set.go
```diff
@@ -1,6 +1,9 @@
 package pdp
```
vasili-v: All the same comments as above are applicable here.

pdp/rule.go
```diff
@@ -1,6 +1,13 @@
 package pdp
```
vasili-v: And here.

mingkaic:
Yes, MarshalWithDepth is better. I'm make sure to name the path marshalling function with the same convention: MarshalPath

pdp/policy.go
```diff
@@ -243,6 +247,52 @@ func (p *Policy) delChild(ID string) (*Policy, error) {
 	return p.updatedCopy(rules, algorithm), nil
 }
 
+// MarshalWithDepth implements StorageMarshal
+func (p Policy) MarshalWithDepth(out io.Writer, depth int) error {
+	if depth < 0 {
+		return newMarshalInvalidDepthError(depth)
+	}
+	pjson, err := json.Marshal(storageNodeFmt{
```
vasili-v: It's better to make a function here like marshalHeader.
```go
func marshalHeader(v interface{}, w io.Writer) err {
	b, err := json.Marshal(v)
	if err != nil {
		return err
	}

	if len(b) < 1 || b[len(b)-1] != '}' {
		return newInvalidHeaderError()
	}

	_, err = w.Write(b[:len(b)-1])
	return err
}
```
And call it here and below.

pdp/policy.go
```diff
@@ -243,6 +247,52 @@ func (p *Policy) delChild(ID string) (*Policy, error) {
...
+		Ord: p.ord,
+		ID:  p.id,
+	})
+	if err != nil {
+		return bindErrorf(err, "pid=\"%s\"", p.id)
+	}
+	_, err = out.Write(append(pjson[:len(pjson)-1],
```
vasili-v: Append here can lead to additional allocation. It's better to rely on writer ability to buffer and just make additional call to `out.Writer([]byte(",\"rules\":["))` instead of trying to buffer here.

pdp/policy.go
vasili-v: I see that [io.Writer description](https://golang.org/pkg/io/#Writer) says: "Write must return a non-nil error if it returns n < len(p)". So we can rely just on error returned from Write.

PR #165, opened 2018-04-19
Write response if external resolver returns SERVFAIL

UladzimirTrehubenka:


contrib/coredns/policy/policy.go
```diff
@@ -702,6 +704,8 @@ Exit:
 
 	if status != dns.RcodeServerFailure {
```
vasili-v:
```go
if status != dns.RcodeServerFailure || resolveFailed {
	w.WriteMsg(r)
}
```

PR #168, opened 2018-04-25
Add Storage Query Endpoint

mingkaic:


pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	depthStr := depthOpt[0]
+	depth, err := strconv.ParseInt(depthStr, 10, 64)
+	if err != nil {
+		io.WriteString(w, err.Error())
+		return
+	}
+
+	// parse path
+	path := strings.FieldsFunc(r.URL.Path, func(c rune) bool { return c == '/' })[1:]
```
vasili-v: Why not `strings.Split(r.URL.Path, "/")`?

pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,49 @@
...
+type storageHandler struct {
+	s *Server
+}
+
+func (handler *storageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// sanity check
+	root := handler.s.p
+	if root == nil {
+		io.WriteString(w, "Server missing policy storage")
+		return
```
vasili-v: You need to return some http error code here (like 404) with `w.WriteHeader` function.

pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// parse depth
+	queryOpt := r.URL.Query()
+	depthOpt, ok := queryOpt["depth"]
+	if !ok {
+		io.WriteString(w, "depth option not specified")
```
vasili-v: Absence of "depth" is not an error. You can safely use 0.

pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,49 @@
...
+		return
+	}
+	depthStr := depthOpt[0]
+	depth, err := strconv.ParseInt(depthStr, 10, 64)
+	if err != nil {
+		io.WriteString(w, err.Error())
+		return
```
vasili-v: Here you need to set bad request status.

pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// parse path
+	path := strings.FieldsFunc(r.URL.Path, func(c rune) bool { return c == '/' })[1:]
+	target, err := root.GetAtPath(path)
+	if err != nil {
+		io.WriteString(w, err.Error())
+		return
```
vasili-v: Here you need to analyze error returned - if it is pathNotFoundError you need to return 404. Otherwise 500. To be able to do so you need to export pathNotFoundError using uppercase first character in errors.yaml.

pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// dump
+	if err = target.MarshalWithDepth(w, int(depth)); err != nil {
+		io.WriteString(w, err.Error())
+		return
```
vasili-v: Error code here 500.

pdpserver/server/ctrl_storage.go
mingkaic: I wasn't sure if http would properly parse out dangling '/'. Example: "127.0.0.1:8080/query/abc/def/?depth=12" => path = "query/abc/def/" => ["query", "abc", "def", ""]. FieldsFunc drops empty strings

mingkaic: On second thought, I'll just use Split, since dangling / can just be a 400 error

vasili-v: Ok!

pdpserver/server/ctrl_storage.go
```diff
@@ -0,0 +1,60 @@
...
+	var (
+		depth int64
+		err   error
+	)
+
+	// parse depth
+	queryOpt := r.URL.Query()
+	if depthOpt, ok := queryOpt["depth"]; ok {
+		depthStr := depthOpt[0]
+		depth, err = strconv.ParseInt(depthStr, 10, 64)
```
vasili-v: We don't want to get negative depth. Please use `strconv.ParseUint`. However uint64 → int can overflow int. I'd put check `if depth > math.MaxInt32 { return 400 }`.

mingkaic: I think there's already a negative depth check in place:
`if depth < 0 { return newMarshalInvalidDepthError(depth) }`

However, you raise a good point. Why not make depth in MarshalWithDepth typed as uint?
We can also use strconv.ParseUint(depthStr, 10, 32) to enforce 32 bit length when parsing.
Do you forsee any problems with this strategy?

vasili-v: Yes, we check the depth at library level but it leads to 500 error while semantically it's 400. You need to export MarshalInvalidDepthError and check for it or put the validation I wrote about.

`strconv.ParseUint(depthStr, 10, 32)` looks good to me.

PR #173, opened 2018-05-10
Integrated with optimized go-trees package

vasili-v:


pdp/content_test.go
```diff
@@ -617,7 +617,7 @@ func TestLocalContentStorage(t *testing.T) {
 	} else {
 		path := []Expression{
 			MakeStringValue("key"),
-			MakeDomainValue(domain.WireNameLower("\x07example\x03gov\x00"))}
```
yuewko: So, we don't support "wire" format anymore?

vasili-v: Yep. Actually, it ins't big deal to add it too. But I decided to not do it as anyway wire require some transformations like changing case and order of labels.

yuewko: OK. Thanks.

vasili-v:
Updated go-trees package

PR #175, opened 2018-05-29
Added decision cache

vasili-v:


vasili-v:
The PR contains preview of changes for adding decision cache for now.

vasili-v:
This is now stable version which is ready for review and I don't expect global changes from now.

vasili-v:
One more thing is still missing - options to turn on/off and control cache in Corefile.

pdp/errors.yaml
```diff
@@ -988,3 +991,417 @@ errors:
...
+
+- id: responseEffectError
+  fields:
+  - id: effect
+    type: int
+  msg: "Unknown effect %d"
+  args:
+  - field: effect
+
+- id: ResponseServerError
```
yuewko: The id `ReponseServerError` should be lowercase.
As it is, the id `ResponseServerErrorID` injected into errors.go becomes exported.

vasili-v: It's intentional. So code which uses the library can distinct the error by type cast and/or by checking its id.

yuewko: Oh, yes, of course!

So, we do this on a case by case basis as the need arises? Or is there a convention that I fail to see?

vasili-v: It's case by case when there is specific need to make this or the need is foreseeable. For the case of `ReponseServerError`, someone may need to know if error returned from PDP or generated locally by for example inability to transfer data.

Additionally, we have a plan to review and make error handling more formal grounded.

yuewko: Perfect!
Exporting error type on a case by case basis is not ideal, IMHO.
Thanks!

pdp/request_test.go
```diff
@@ -0,0 +1,2093 @@
...
+
+	n, err = MarshalRequestAssignments([]byte{}, testRequestAssignments)
+	assertRequestBufferOverflow(t, "MarshalRequestAssignments", err, n)
+
+	n, err = MarshalRequestAssignments(b[:2], testRequestAssignments)
+	if err == nil {
+		t.Errorf("expected no data put to small buffer but got %d", n)
+	} else if _, ok := err.(*requestBufferOverflowError); !ok {
+		t.Errorf("expected *requestBufferOverflowError but got %T (%s)", err, err)
+	}
```
yuewko: It looks to me `assertRequestBufferOverflow()` used in L42 does exactly what L45-L49 does. Is there any reason why we don't reuse `assertRequestBufferOverflow()` here?

There seems to be other occurrences following too.

vasili-v: Nice catch!

PR #177, opened 2018-06-11
Tight integration rebase

yuewko:
This is a WIP for the use of github.com/coredns/policy, the goal of which is to create a CoreDNS policy plugin that process policy request using a linked in version of Themis.


vasili-v:
I thought about the option to have client local server too. But I'd rather call it "built-in" instead of integrated. And it doesn't need to be server (so server shouldn't be touched by the change). It should be just library integrated to client (so only pep changes are needed). In the case you can skip marshaling if user want to use PDP provided primitives to communicate with library. But you need to introduce specific marshalers for reflection part (which is main challenge and can be left for future).

yuewko:
@vasili-v , thanks for your review!

> I thought about the option to have client local server too. But I'd rather call it "built-in" instead of integrated.

The original concept is that this allows the policy plugin to *integrate* PDP into CoreDNS. But, "built-in" works too!

>  And it doesn't need to be server (so server shouldn't be touched by the change). It should be just library integrated to client (so only pep changes are needed).

Sounds good. I'll give that a shot.

> In the case you can skip marshaling if user want to use PDP provided primitives to communicate with library. But you need to introduce specific marshalers for reflection part (which is main challenge and can be left for future).

Yes! This has already been discussed. @rajansandeep will be implementing such improvements.

yuewko:
>>I thought about the option to have client local server too. But I'd rather call it "built-in" instead of integrated.

>The original concept is that this allows the policy plugin to integrate PDP into CoreDNS. But, "built-in" works too!

I have renamed it to `builtinClient`.

yuewko:
>> And it doesn't need to be server (so server shouldn't be touched by the change). It should be just library integrated to client (so only pep changes are needed).

> Sounds good. I'll give that a shot.

Revisiting the code, the reason why I utilized the `Server` object is that the plan is for the builtin PDP to still support, for example, the Control Endpoint to allow policies and contents being pushed from a PAP. Again, this is still very much a WIP.

So, for now, I've only renamed Integrated Server to Builtin Server.

Please let me know what you think.

vasili-v:
Let me think of the control endpoint. It looks more like we need different servers for control and validate so you can embed control server without other stuff.

pep/builtin_client.go
```diff
@@ -0,0 +1,42 @@
...
+	"golang.org/x/net/context"
+
+	ps "github.com/infobloxopen/themis/pdpserver/server"
+)
+
+type builtinClient struct {
+	s *ps.Server
+}
+
+func NewBuiltinClient(policyFile string, contentFiles []string) *builtinClient {
```
vasili-v: Instead of making separate `NewBuiltinClient`. You can provide option `WithBuiltinServer`. So client can be created in common way - `NewClient(WithBuiltinServer(s))`

PR #178, opened 2018-06-13
Added try and concat functions

vasili-v:


pdp/expr_try.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (f functionTry) Calculate(ctx *Context) (AttributeValue, error) {
+	var (
+		v   AttributeValue
+		err error
+	)
+
+	for _, arg := range f.args {
+		v, err = arg.Calculate(ctx)
```
rdrozhdzh: would be nice to check the value and continue if empty

vasili-v: Not every type has "empty" value.

rdrozhdzh: ok, looks good

PR #179, opened 2018-06-13
pdpctrl-client upload check io.EOF error for on Send

mingkaic:
Even though send returning io.EOF error is problematic, we should send all chunks to get more informative error from PDP server

### DIFF
--- a/pdpctrl-client/client.go
+++ b/pdpctrl-client/client.go
@@ -119,7 +119,7 @@ func (c *Client) Upload(id int32, r io.Reader) (int32, error) {
 			chunk := &pb.Chunk{
 				Id:   id,
 				Data: string(p[:n])}
-			if err := u.Send(chunk); err != nil {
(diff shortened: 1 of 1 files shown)


vasili-v:
I'm going to close the PR because to get real error from PDP you need to send not all data but end of transmission mark. The mark can be send by CloseAndRecv function from closeUpload. So correct code could be:
```go
if err := u.Send(chunk); err != nil {
	if err == io.EOF {
		return c.closeUpload(u)
	}
}
```
However `io.EOF` on send occurs when upstream intentionally closes its read end (either PDP or proxy before it). So it can't receive the mark and therefore has no any chance to answer you. And you just get the same `io.EOF` or timeout later. Unless you have example when it's possible. If so create test which shows it. But any way you don't need to send all remaining data.

For some cases you may need to distinct errors from `n, err := r.Read(p)` and `err := u.Send(chunk)`. But it's different thing.

vasili-v:
Closed as per comment above.

PR #180, opened 2018-06-19
Added check for subalgorithm to JAST parser

vasili-v:


huynhmb359:
@vasili-v   I think this should go into janus-2.2 branch.

PR #181, opened 2018-06-19
Added specific logic for MissingValueError to concat expression

vasili-v:


rdrozhdzh:
Perhaps, the doc should be updated

PR #184, opened 2018-06-25
 Added string type as argument for concat expression

rdrozhdzh:


README.md
```diff
@@ -605,7 +605,7 @@ Each of the above numerical functions performs the respective numerical operatio
...
 - **list of strings** - converts its argument to list of strings. It accepts set of strings, list of strings and flags. In case of set of strings the function returns list of strings sorted in order maintained by set (set keeps order of initial value definition). List of strings returned by the function as is. And for flags it returns list of names for flags which are set (keeping order of names from flags type definition).
-- **concat** - concatenates all given arguments to single list of strings. The function treats MissingValueError in special way. If at least one argument returns some data, any MissingValueError is ignored. But when all arguments return the error, **concat** returns the error as well. It accepts lists of strings, sets of strings and flags as arguments similarly to **list of strings**.
+- **concat** - concatenates all given arguments to single list of strings. The function treats MissingValueError in special way. If at least one argument returns some data, any MissingValueError is ignored. But when all arguments return the error, **concat** returns the error as well. It accepts strings, lists of strings, sets of strings and flags as arguments.
```
yuewko: I assume the reason you removed `similarly to **list of strings**` is because **list of strings** above does not accept strings.

However, I think you need a sentence to say that **concat** handles lists of strings, sets of strings and flags the same way as **list of strings**.

PR #186, opened 2018-06-29
Marshal Targets, Obligations, and Algorithms

mingkaic:
I added MarshalJSON functions to the algorithms to easily marshal evaluables.
I'm under the assumption that recursive algorithms like Mapper wouldn't be very deep and therefore wouldn't have huge memory impact when marshalled to string

### DIFF
--- a/pdp/flags_mapper_pca.go
+++ b/pdp/flags_mapper_pca.go
@@ -63,8 +63,13 @@ func (a flagsMapperPCA) execute(policies []Evaluable, ctx *Context) Response {
 }
 
 func (a flagsMapperPCA) MarshalJSON() ([]byte, error) {
-	defID, _ := a.def.GetID()
-	errID, _ := a.err.GetID()
+	var defID, errID string
+	if a.def != nil {
+		defID, _ = a.def.GetID()
+	}
+	if a.err != nil {
+		errID, _ = a.err.GetID()
+	}
 	return json.Marshal(mapperAlgFmt{
 		Type:      "flagsMapperPCA",
 		Default:   strconv.Quote(defID),
--- a/pdp/flags_mapper_pca.go
+++ b/pdp/flags_mapper_pca.go
@@ -1,6 +1,10 @@
 package pdp
 
-import "sort"
+import (
+	"encoding/json"
+	"sort"
+	"strconv"
+)
 
 type flagsMapperPCA struct {
 	argument  Expression
@@ -56,6 +60,17 @@ func (a flagsMapperPCA) execute(policies []Evaluable, ctx *Context) Response {
 	}
 
 	return a.calculateErrorPolicy(ctx, newFlagsMapperRCAArgumentTypeError(t))
+}
+
+func (a flagsMapperPCA) MarshalJSON() ([]byte, error) {
+	defID, _ := a.def.GetID()
+	errID, _ := a.err.GetID()
+	return json.Marshal(mapperAlgFmt{
+		Type:      "flagsMapperPCA",
+		Default:   strconv.Quote(defID),
+		Error:     strconv.Quote(errID),
+		Algorithm: a.algorithm,
+	})
 }
 
 func (a flagsMapperPCA) add(ID string, child, old Evaluable) PolicyCombiningAlg {
--- a/pdp/flags_mapper_pca_test.go
+++ b/pdp/flags_mapper_pca_test.go
@@ -1,6 +1,9 @@
 package pdp
 
-import "testing"
+import (
+	"encoding/json"
+	"testing"
+)
 
 func TestFlagsMapperPCAOrdering(t *testing.T) {
 	ft, err := NewFlagsType("flags", "third", "first", "second")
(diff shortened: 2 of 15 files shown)


pdp/flags_mapper_pca.go
```diff
@@ -58,6 +62,17 @@ func (a flagsMapperPCA) execute(policies []Evaluable, ctx *Context) Response {
 	return a.calculateErrorPolicy(ctx, newFlagsMapperRCAArgumentTypeError(t))
 }
 
+func (a flagsMapperPCA) MarshalJSON() ([]byte, error) {
+	defID, _ := a.def.GetID()
```
vasili-v: Both `.def` and `.err` can be nil so you get segfault here.

pdp/flags_mapper_rca.go
```diff
@@ -60,6 +62,15 @@ func (a flagsMapperRCA) execute(rules []*Rule, ctx *Context) Response {
 	return a.calculateErrorRule(ctx, newFlagsMapperRCAArgumentTypeError(t))
 }
 
+func (a flagsMapperRCA) MarshalJSON() ([]byte, error) {
+	return json.Marshal(mapperAlgFmt{
+		Type:      "flagsMapperRCA",
+		Default:   strconv.Quote(a.def.id),
```
vasili-v: The same issue.

pdp/mapper_pca.go
```diff
@@ -177,6 +179,17 @@ func makeMapperPCA(policies []Evaluable, params interface{}) PolicyCombiningAlg
 		algorithm: mapperParams.Algorithm}
 }
 
+func (a mapperPCA) MarshalJSON() ([]byte, error) {
+	defID, _ := a.def.GetID()
```
vasili-v: Again.

pdp/flags_mapper_pca_test.go
```diff
@@ -94,3 +97,58 @@ func TestFlagsMapperPCAOrdering(t *testing.T) {
 		}
 	}
 }
+
+const (
+	expectEmptyFlagMapperPCAJSON = `{"type":"flagsMapperPCA","def":"\"\"","err":"\"\"","alg":{"type":"firstApplicableEffectPCA"}}`
+	expectFlagMapperPCAJSON      = `{"type":"flagsMapperPCA","def":"\"first\"","err":"\"second\"","alg":{"type":"firstApplicableEffectPCA"}}`
+)
```
yuewko: Since these constants are only used in the func below, I would move them inside the func.

pdp/flags_mapper_pca_test.go
```diff
@@ -94,3 +97,58 @@ func TestFlagsMapperPCAOrdering(t *testing.T) {
...
+	algParam := MapperPCAParams{
+		Argument:  AttributeDesignator{a: Attribute{id: "f", t: ft}},
+		Order:     MapperPCAInternalOrder,
+		Algorithm: firstApplicableEffectPCA{},
+	}
+	alg := makeMapperPCA(policies, algParam)
+	b, err := json.Marshal(alg)
+	if err != nil {
+		t.Fatalf("Expected no marshaling error but got: %s", err)
+	} else if expectEmptyFlagMapperPCAJSON != string(b) {
```
yuewko: I would do `string(b) != expectEmptyFlagMapperPCAJSON` instead.
This is no different from `err != nil` a couple of lines above.

pdp/flags_mapper_pca_test.go
```diff
@@ -94,3 +97,58 @@ func TestFlagsMapperPCAOrdering(t *testing.T) {
...
+		Def:       "first",
+		DefOk:     true,
+		Err:       "second",
+		ErrOk:     true,
+	}
+	alg = makeMapperPCA(policies, algParam2)
+	b, err = json.Marshal(alg)
+	if err != nil {
+		t.Fatalf("Expected no marshaling error but got: %s", err)
+	} else if expectFlagMapperPCAJSON != string(b) {
```
yuewko: ditto

mingkaic:
I've address ko's concerns
@vasili-v can you review again? Thanks

pdp/flags_mapper_pca_test.go
```diff
@@ -128,10 +128,8 @@ func TestFlagsMapperPCAMarshal(t *testing.T) {
 		Algorithm: firstApplicableEffectPCA{},
 	}
 	alg := makeMapperPCA(policies, algParam)
-	b, err := json.Marshal(alg)
-	if err != nil {
-		t.Fatalf("Expected no marshaling error but got: %s", err)
-	} else if expectEmptyFlagMapperPCAJSON != string(b) {
+	b, _ := json.Marshal(alg)
+	if expectEmptyFlagMapperPCAJSON != string(b) {
```
yuewko: Sorry if my original comment was misleading..

There's nothing wrong with the original code:
```
b, err := json.Marshal(alg)
if err  != nil {
```
etc

The main point is:
```
if expectEmptyFlagMapperPCAJSON != string(b) { 
```
should be:
```
if string(b) != expectEmptyFlagMapperPCAJSON  { 
```
instead.

PR #187, opened 2018-07-07
plugin/policy: fix race in AttrGauge

rdrozhdzh:
- syncronize AddAttributes() with a channel
 - use single global AttrGauge across all instances of policy plugin

### DIFF
--- a/contrib/coredns/policy/setup.go
+++ b/contrib/coredns/policy/setup.go
@@ -6,7 +6,6 @@ import (
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/dnstap"
-	"github.com/coredns/coredns/plugin/metrics"
 
 	"github.com/mholt/caddy"
 )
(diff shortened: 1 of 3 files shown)


contrib/coredns/policy/setup.go
```diff
@@ -38,19 +37,11 @@ func setup(c *caddy.Controller) error {
...
-				})
-			}
-		}
-		return nil
+		return policyPlugin.SetupMetrics(c)
 	})
 
 	c.OnShutdown(func() error {
 		policyPlugin.closeConn()
-		policyPlugin.attrGauges.Stop()
```
yuewko: I can see why we don't call `AttrGuage.Stop()` from here anymore.

However, it now looks like the `AttrGuage.Stop()` is not used at all, except in UTs.

Is that intentional?

rdrozhdzh: That's right, now the global AttrGauge (if started) will run until the coredns process is terminated

yuewko: Thanks for the confirmation!

I'm going to approve it. I suggest you make a comment in the code (maybe where Start() is called) to that effect so it's clear.

PR #189, opened 2018-07-10
Missing symbols when initializing LocalContent

yuewko:
In LocalContentStorageTransaction.Commit(), symbols was missing
when LocalContent is initialized.


vasili-v:
It would be good to have a test for the case.

yuewko:
```
It would be good to have a test for the case.
```
@vasili-v , I'll do so shortly as a separate PR.
Thanks for the review.

PR #195, opened 2018-07-24
Compare Domain IP Response Rule Precedence in Coredns plugin

mingkaic:


contrib/coredns/policy/policy.go
```diff
@@ -163,11 +163,31 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
 				attrsResponse = p.attrPool.Get()
 				defer p.attrPool.Put(attrsResponse)
 			}
+			queryAction, queryRedirect := ah.action, ah.dst
 			// validate response IP (validation #2)
 			if err := p.validate(ah, attrsResponse); err != nil {
 				status = dns.RcodeServerFailure
 				return dns.RcodeSuccess, err
 			}
+			var qOrder, rOrder string
```
vasili-v: This code makes assumption on how policy looks. We can't do that.

PR #199, opened 2018-08-01
POC: Sharding

vasili-v:
This is just an example of a way to implement sharding. Production implementation needs more solid design, to be cleaned up of shortcuts, to get more validations of inputs and more test coverage.


yuewko:
LGTM

PR #200, opened 2018-08-03
policy plugin: always resolve domain

rdrozhdzh:
- resolve domain even if validation gives block or redirect action,
   because IP validation policy may have higher priority


rdrozhdzh:
Closing this in favour of PR #212

PR #204, opened 2018-08-08
Added support for formatting the mac address.

masumbuet:
This will format the mac address from "0cc47a507753" to "0c:c4:7a:50:77:53" as asked in JANUS-16336.


huynhmb359:
@masumbuet   Since the fix will be done on TI.   Could you please close this PR?

PR #205, opened 2018-08-12
POC: PEP Client Validate return Cached flag

yuewko:
- Added pep.CachedResponse that encapsulate pdp.Response with
  the addition of a Cached flag. The default pep unary and
  streaming client updates Cached flag if a CacheResponse is
  specified.
- Updated CoreDNS policy plugin to use pep.CachedResponse
- (somewhat related)Updated CoreDNS policy plugin to use CoreDNS
  logger.

### DIFF
--- a/contrib/coredns/policy/setup.go
+++ b/contrib/coredns/policy/setup.go
@@ -4,9 +4,11 @@ import (
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/dnstap"
-
+	clog "github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/mholt/caddy"
 )
+
+var log = clog.NewWithPlugin("policy")
 
 func init() {
 	caddy.RegisterPlugin("policy", caddy.Plugin{
--- a/pep/unmarshal.go
+++ b/pep/unmarshal.go
@@ -37,29 +37,15 @@ var (
 	}
 )
 
-func fillResponse(res pb.Msg, v interface{}) error {
-	switch v := v.(type) {
-	case *pb.Msg:
-		*v = res
-		return nil
+type CachedResponse struct {
+	pdp.Response
 
-	case *pdp.Response:
-		if v.Obligations == nil {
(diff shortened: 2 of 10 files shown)


contrib/coredns/policy/setup.go
```diff
@@ -4,10 +4,12 @@ import (
 	"github.com/coredns/coredns/core/dnsserver"
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/dnstap"
-
+	clog "github.com/coredns/coredns/plugin/pkg/log"
 	"github.com/mholt/caddy"
 )
 
+var log = clog.NewWithPlugin("policy")
```
vasili-v: Let's change the way plugin works with logs as separate PR.

pep/unmarshal.go
```diff
@@ -37,29 +37,15 @@ var (
...
-
-	case *pdp.Response:
-		if v.Obligations == nil {
-			effect, a, err := pdp.UnmarshalResponseAssignments(res.Body)
-			if err != nil {
-				if _, ok := err.(*pdp.ResponseServerError); !ok {
-					return err
-				}
-			}
+type CachedResponse struct {
```
vasili-v: The name `CachedResponse` tells that response goes from cache. I'd name it `ResponseEx` (as MS Windows usually does for structures with extra flags).

contrib/coredns/policy/setup.go
yuewko: Yup, cannot agree more.

pep/unmarshal.go
yuewko: Seem to recall seeing those when I last work with the WIN32 API - some 20 years ago...

From your earlier review comment though, we don't need to introduce a new struct anyway.

yuewko:
@vasili-v, I'll submit two PRs based on your comments. Thanks!!

pep/unmarshal.go
vasili-v: 😀

yuewko:
Address logging change comment via #206.

vasili-v:
Closing as per comment above another solution has been implemented.

PR #206, opened 2018-08-13
Updated CoreDNS plugin to use CoreDNS logger

yuewko:
Updating to use CoreDNS logger result in the following changes in log messages generated by the CoreDNS policy plugin:

- The log message will now include the name of the plugin, which is included in the log message after the log level label. For example:
From:
`2018/08/13 09:29:07 [INFO] PDP request: [(string)type:query (domain)domain_name:www.yahoo.com. (string)dns_qtype:1 (address)source_ip:127.0.0.1]`
To:
`2018/08/13 09:31:32 [INFO] plugin/policy: PDP request: [(string)type:query (domain)domain_name:www.yahoo.com. (string)dns_qtype:1 (address)source_ip:127.0.0.1]`
 
- In order to NOT have to enable debug for CoreDNS and all its plugins, a number of DEBUG messages is now changed to INFO messages. This is to maintain existing behavior - currently, policy plugin DEBUG messages are log.Printf()s and are therefore always output, however, with the CoreDNS logger, DEBUG messages are only output if debug mode is turned on.

- Warning messages will changed from being labelled as WARN to WARNING

### DIFF
--- a/contrib/coredns/policy/attrholder.go
+++ b/contrib/coredns/policy/attrholder.go
@@ -2,7 +2,6 @@ package policy
 
 import (
 	"fmt"
-	"log"
 	"net"
 	"strconv"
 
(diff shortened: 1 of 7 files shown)


yuewko:
@vasili-v , @rdrozhdzh: I have added a commit that changes some DEBUG to INFO. I have also updated the comment to describe the detail of the changes.

PTAL. Thanks!!

PR #207, opened 2018-08-13
Added WithOnCacheHitHandler PEP Option

yuewko:
The option allows a handler func to be specified which
is called when a decision request is served from the
decision cache.

Also updated contrib/coredns/policy to make use of this to
log decision request processed via decision cache.

### DIFF
--- a/contrib/coredns/policy/client.go
+++ b/contrib/coredns/policy/client.go
@@ -11,7 +11,7 @@ import (
 
 type pepCacheHitHandler struct{}
 
-func (ch *pepCacheHitHandler) Handle(req interface{}, resp interface{}) {
+func (ch *pepCacheHitHandler) Handle(req interface{}, resp interface{}, err error) {
 	log.Printf("[INFO] PEP responding to PDP request from cache %+v", req)
 }
 
--- a/contrib/coredns/policy/client.go
+++ b/contrib/coredns/policy/client.go
@@ -9,6 +9,16 @@ import (
 	"github.com/infobloxopen/themis/pep"
 )
 
+type pepCacheHitHandler struct{}
+
+func (ch *pepCacheHitHandler) Handle(req interface{}, resp interface{}) {
+	log.Printf("[INFO] PEP responding to PDP request from cache %+v", req)
+}
+
+func newPepCacheHitHandler() *pepCacheHitHandler {
+	return &pepCacheHitHandler{}
+}
+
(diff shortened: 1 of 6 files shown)


pep/streaming_client.go
```diff
@@ -134,8 +134,15 @@ func (c *streamingClient) Validate(in, out interface{}) error {
 	}
 
 	if c.cache != nil {
-		if b, err := c.cache.Get(string(m.Body)); err == nil {
-			return fillResponse(pb.Msg{Body: b}, out)
+		var b []byte
+		if b, err = c.cache.Get(string(m.Body)); err == nil {
+			err = fillResponse(pb.Msg{Body: b}, out)
+			if err == nil {
```
vasili-v: I think that the handler should be called anyway even error on unmarshalling occurs. And it should take error as well:
```go
err = fillResponse(pb.Msg{Body: b}, out)
if c.opts.onCacheHitHandler != nil {
	c.opts.onCacheHitHandler.Handle(in, out, err)
}
```
It needs to tell user that we hit cache anyway even if we can't unmarshal cached response. We even can add more complicated logic to return raw response in case of error:
```go
err = fillResponse(pb.Msg{Body: b}, out)
if c.opts.onCacheHitHandler != nil {
	if err != nil {
		c.opts.onCacheHitHandler.Handle(in, b, err)
	} else {
		c.opts.onCacheHitHandler.Handle(in, out, nil)
	}
}
```

yuewko:
@vasili-v , updated PR as per your comment.

I haven't updated the handler in the policy plugin to handle the error case. I believe we should keep the current behavior unless CoreDNS is in debug mode. I'll add debug log after #206 is merged.

PTAL. Thanks!

contrib/coredns/policy/client.go
```diff
@@ -9,6 +9,16 @@ import (
 	"github.com/infobloxopen/themis/pep"
 )
 
+type pepCacheHitHandler struct{}
+
+func (ch *pepCacheHitHandler) Handle(req interface{}, resp interface{}, err error) {
+	log.Printf("[INFO] PEP responding to PDP request from cache %+v", req)
```
rdrozhdzh: This line breaks master. Should be log.Infof()

yuewko: Yes, that was my bad. 

I've already submitted PR #207 to fix that. PTAL. 

Thanks!

PR #212, opened 2018-09-06
policy plugin: refactor attribute holder

rdrozhdzh:
- added/modified attribute configuration in corefile. Affected options
   are validation1, validation2, default_decision, metrics, dnstap. See
   more details in README.md

 - fallback to default decision in case if PDP validation failed.

 - refactored debug query

 - added integer "policy_action" attribute which contains explicit
   action, i.e. now action is not deduced from existence and/or content of
   other attributes. Removed attributes "refuse" and "drop"

 - attribute "log" now defines the dnstap log level, i.e. the set of
   attributes to be sent with dnstap message

### DIFF
--- a/contrib/coredns/policy/README.md
+++ b/contrib/coredns/policy/README.md
@@ -1,94 +1,102 @@
 # policy
 
-*policy* is the Policy Enforcement Point for Themis project implemented as CoreDNS plugin.
+*policy* is the Policy Enforcement Point (PEP) for Themis project implemented as CoreDNS plugin.
 
 ## Syntax
 
 ~~~ txt
 policy {
-    endpoint ADDR_1, ADDR_2, ... ADDR_N
-    edns0 CODE NAME [SRCTYPE DSTTYPE] [SIZE START END]
-    ...
-    debug_query_suffix SUFFIX
-    debug_id ID
+    endpoint ADDR_1 ADDR_2 ... ADDR_N
     streams COUNT
-    transfer ATTR_1, ATTR_2, ... ATTR_N
-    dnstap ATTR_1, ATTR_2, ... ATTR_N
-    metrics ATTR_1, ATTR_2, ... ATTR_N
-    passthrough SUFFIX_1, SUFFIX_2, ... SUFFIX_N
-    log
+    connection_timeout TIMEOUT
     max_request_size [[auto] SIZE]
     max_response_attributes auto | COUNT
     cache [TTL [SIZE]]
+
+    edns0 CODE NAME [SRCTYPE] [SIZE START END]
+    validation1 ATTR_1 ATTR_2 ... ATTR_N
+    validation2 ATTR_1 ATTR_2 ... ATTR_N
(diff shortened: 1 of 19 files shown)


rdrozhdzh:
Hm, do we still support go 1.9?

contrib/coredns/policy/README.md
```diff
@@ -1,94 +1,102 @@
...
 
-Option debug_query_suffix SUFFIX (should have dot at the end) enables debug query feature.
+Option **edns0** is used for parsing edns0 options into PDP attributes, option with code *CODE* is parsed as attribute with name *NAME*. *CODE* can be defined as octal, decimal or hexadecimal value. Hexadecimal numbers should start with prefix `0x`, e.g `0xfff5`, octal numbers should start with `0`, e.g. `0177765`. *SRCTYPE* defines encoding if edns0 data, valid values are `hex` (default), `bytes`, `ip`. Params *SIZE*, *START*, *END* is supported only for *SRCTYPE* = `hex`. Setting param *SIZE* to value > 0 enables edns0 option data size check. Param *START* and *END* (last data byte index + 1) allows picking out a particular part of edns0 option data into a separate attribute. Option **edns0** can be used repeatedly to define several ends0 attributes
 
-Option debug_id set string that is used for debug query response as unique id for determine what CoreDNS instance replies on the request.
+Option **validation1** defines a set of attributes to be sent to PDP for validation before resolving domain name.
```
yuewko: Since this is specifically applied to *validation before resolving domain name*, why don't we call this option, say, `validation_domain`, instead of `validation1`?

contrib/coredns/policy/README.md
```diff
@@ -1,94 +1,102 @@
...
 
-Option streams set gRPC streams count for PDP connection.
+Option **validation2** defines a set of attributes to be sent to PDP for validation after resolving domain name.
```
yuewko: ditto: e.g. `validate_ip`

contrib/coredns/policy/README.md
```diff
@@ -1,94 +1,102 @@
...
-    dnstap pid
-    metrics uid
+    connection_timeout 1s
+
+    edns0 0xffee client_id hex 32 0 32
+    edns0 0xffee group_id hex 32 16 32
+    edns0 0xffef uid
+    edns0 0xffea client_ip ip
+    edns0 0xffeb client_name bytes
+    validation1 type="query" domain_name
```
yuewko: I understand that, to specify a string like `query`, it *must be* quoted. On the other hand, to specify an integer or IP address, it must *not* be quoted.

I believe this should be made clear in the documentation earlier

contrib/coredns/policy/README.md
rdrozhdzh: User can specify *ANY* attributes for validation1 and validation2, so calling them validate_domain, validate_ip may be confusing. Moreover, user may want to skip validation1 and pass both "domain_name" and "address" with validation2, i.e. make a single validation request to PDP

PR #218, opened 2018-09-21
[Preview] Unified PIP

vasili-v:
This is open source part of Unified PIP which covers:
- server and client libraries implementing fast TCP/Unix protocol with low memory and GC usage;
- CLI client to test and profile server;
- example of server implementation similar to local selector.


pip/server/options.go
```diff
@@ -0,0 +1,28 @@
+package server
+
+// Option configures how we set up PDP server.
```
rdrozhdzh: PIP server

pip/server/options.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type options struct {
+	net  string
+	addr string
+}
+
+var defaults = options{
+	net:  "tcp",
+	addr: "localhost:5555",
```
rdrozhdzh: consider changing default port to avoid possible conflict with PDP

pip/server/server.go
```diff
@@ -0,0 +1,92 @@
...
+// Bind links server to a port or file.
+func (s *Server) Bind() error {
+	if s.ln != nil {
+		return ErrBound
+	}
+
+	nw := strings.ToLower(s.opts.net)
+	if nw == "unix" {
+		if err := os.Remove(s.opts.addr); err != nil {
+			if pErr, ok := err.(*os.PathError); !ok || pErr.Err != syscall.ENOENT {
```
rdrozhdzh: please re-check the condition, perhaps you meant to use && instead of ||.

pip/server/server_test.go
```diff
@@ -0,0 +1,180 @@
...
+		WithAddress("127.0.0.1:0"),
+	)
+
+	if assert.NoError(t, s.Bind()) {
+		assert.Equal(t, ErrBound, s.Bind())
+	}
+}
+
+func TestServerBindInvalidNetwork(t *testing.T) {
+	assert.NotEqual(t, nil, NewServer(WithNetwork("invalid")).Bind())
```
rdrozhdzh: consider splitting the code into several lines like other tests to make the tests style consistent and the code clearer

pip/server/server.go
vasili-v: It's correct. We ignore error when `ok && pErr.Err == syscall.ENOENT`.

rdrozhdzh: ok, this way is clearer

pip/server/options.go
```diff
@@ -0,0 +1,90 @@
...
+	"math"
+	"net"
+	"time"
+)
+
+// Option configures how we set up PIP server.
+type Option func(*options)
+
+// WithNetwork returns an Option which sets service network. It supports "tcp", "tcp4", "tcp6" and "unix" netwroks.
+func WithNetwork(net string) Option {
```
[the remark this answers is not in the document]
vasili-v: Package net calls this thing network so I just keep their naming.

pip/server/reader.go
```diff
@@ -0,0 +1,87 @@
...
+	"errors"
+	"io"
+)
+
+// ErrMsgOverflow indicates that message is too big for message buffer.
+var ErrMsgOverflow = errors.New("message buffer overflow")
+
+const msgSizeBytes = 4
+
+func read(c connWithErrHandler, msgs pool, bufSize int) chan []byte {
```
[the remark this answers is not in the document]
vasili-v: Yes. It's one of the point of the work. gRPC much slower and makes a lot of allocations (actually, it's slower because of the allocations). I made quick benchmark and was able to create zero allocations server based on TCP/Unix networks. And we already have serialization functions as part of PDP.

pip/server/options.go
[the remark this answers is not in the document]
yuewko: Ok

pip/server/server.go
```diff
@@ -0,0 +1,158 @@
...
+
+	ln := s.ln
+	s.ln = nil
+
+	return ln.Close()
+}
+
+const netListenerClosedMsg = "use of closed network connection"
+
+func isLisenerClosed(err error) bool {
```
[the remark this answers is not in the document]
yuewko: Looks like it's been renamed.

pip/server/reader.go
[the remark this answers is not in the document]
vasili-v: Yes, this is the plan. As a result of the work you need to implement only server PIP handler according business logic you want. While transport and balancing implemented by the package.

pip/client/client.go
```diff
@@ -0,0 +1,229 @@
...
+	// Connect establishes connection to given PIP server.
+	Connect() error
+
+	// Close terminates previously established connection if any. Close silently
+	// returns if connection hasn't been established yet or if it has been
+	// already closed.
+	Close()
+
+	// Get requests information from PIP.
+	Get(args []pdp.AttributeAssignment) (pdp.AttributeValue, error)
```
yuewko: Would you consider returning `[]pdp.AttributeAssignment` instead of `pdp.AttributeValue`? It seems somewhat restricting as it stands.

pip/client/k8s.go
```diff
@@ -0,0 +1,67 @@
+package client
```
yuewko: The discovery code looks very general - i.e. not pip client specific. Would you consider making this a separate package so it can potentially be reused elsewhere?

vasili-v: Then it's better to put it in some other github.com/infobloxopen repo.

vasili-v:
I removed "Preview" label from title as no more big changes expected here (except PIP cache which is part of ongoing optimization effort). Please walk through the UPIP feature implementation once again. Your comments and suggestions are welcome!

PR #221, opened 2018-10-31
Add API description to ctrl_storage endpoint & Make AttributeAssignment Marshalable

mingkaic:
I've noticed that the storage control/query endpoint has very few API documentation, so I'm adding the usage description to 404 error message. I've also exposed AttributeAssignment values when marshaling to JSON to get information on obligations.

### DIFF
--- a/pdp/assignment.go
+++ b/pdp/assignment.go
@@ -1,6 +1,7 @@
 package pdp
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 
@@ -16,6 +17,14 @@ import (
 type AttributeAssignment struct {
 	a Attribute
 	e Expression
+}
+
+// AttribAssignFmt is the json marshal format of serialized AttributeAssignment
+type AttribAssignFmt struct {
+	Name      string
+	NameType  string
+	Value     string
+	ValueType string
 }
 
 // MakeAttributeAssignment creates assignment of given expression to given
@@ -458,6 +467,20 @@ func (a AttributeAssignment) String() string {
 		return err.Error()
 	}
 	return fmt.Sprintf("(%s)%s:%s", valueType, name, value)
+}
+
+// MarshalJSON satisfies Marshaler interface
+func (a AttributeAssignment) MarshalJSON() ([]byte, error) {
+	name, valueType, value, err := a.Serialize(nil)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(AttribAssignFmt{
+		Name:      name,
+		NameType:  a.a.t.GetKey(),
+		Value:     value,
+		ValueType: valueType,
+	})
 }
 
 func (a AttributeAssignment) bindError(err error) error {
--- a/pdp/assignment.go
+++ b/pdp/assignment.go
@@ -471,6 +471,8 @@ func (a AttributeAssignment) String() string {
 }
 
 // MarshalJSON satisfies Marshaler interface
+// Only works for assignment expression where righthand doesn't depend on context
+// E.g.: values, constant expression, selector that don't rely on attributes or local content
 func (a AttributeAssignment) MarshalJSON() ([]byte, error) {
 	name, valueType, value, err := a.Serialize(emptyCtx)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pdp/assignment.go
```diff
@@ -18,6 +19,14 @@ type AttributeAssignment struct {
 	e Expression
 }
 
+// AttribAssignFmt is the json marshal format of serialized AttributeAssignment
+type AttribAssignFmt struct {
+	Name      string
+	NameType  string
```
vasili-v: Assignment target (lvalue) and expression to assign (rvalue) must have the same type so it's enough to have only Name, Type and... By the way at right side of the assignment we can have expression of any complexity. So it isn't correct to limit ourselves just with "Value". Here you should have marshalable "Expression" which creates something like this on JSON for `var x bool = s == "test"`:
```json
{
  "Name": "x",
  "Type": "boolean",
  "Value": {
    "equal": [
      {
        "attr": {
          "name": "s",
          "type": "string"
        }
      },
      {
        "value": {
          "type": "string",
          "content": "test"
        }
      }
    ]
  }
}
```

pdp/assignment.go
```diff
@@ -460,6 +469,20 @@ func (a AttributeAssignment) String() string {
 	return fmt.Sprintf("(%s)%s:%s", valueType, name, value)
 }
 
+// MarshalJSON satisfies Marshaler interface
+func (a AttributeAssignment) MarshalJSON() ([]byte, error) {
+	name, valueType, value, err := a.Serialize(nil)
```
vasili-v: Generally, `Serialize` panics in its argument is `nil`. It is a consequence of attempt to represent any expression as value - generally, it isn't possible. For now you can add check that `a.e` is instance of `AttributeValue`. In that case serialization works with no context. For other type you can put something like `"<complicated expression>"` instead of value. Later you need to add `MarshalJSON` to `Expression` interface and to implement it for all structures which implementing `Expression` interface. There are bunch of them starting with `AttributeValue`, `AttributeDesignator`, and so on. Actually, there are so many of them that implementation of marshaling for them is main part of this work.

yuewko: I think it would also be useful, e.g. for diagnostics purposes, to be able to Serialize (at least not panic) even when the Expression cannot be evaluated due to missing `context` - in which case, we could just include whatever information that is available.

mingkaic:
@vasili-v @yuewko Can you guys review this PR again? I've added unittests for the assignment expression json marshalling. To address the nil context problem, I've taken the coredns approach of serializing with an empty context. The tests show this approach works.

pdp/assignment.go
```diff
@@ -453,13 +463,26 @@ func (a AttributeAssignment) Serialize(ctx *Context) (string, string, string, er
...
 func (a AttributeAssignment) String() string {
-	name, valueType, value, err := a.Serialize(nil)
+	name, valueType, value, err := a.Serialize(emptyCtx)
 	if err != nil {
 		return err.Error()
 	}
 	return fmt.Sprintf("(%s)%s:%s", valueType, name, value)
 }
 
+// MarshalJSON satisfies Marshaler interface
```
vasili-v: Please mention here in the comment that it works for an assignment expression which righthand part doesn't depend on context. Examples of such expression are a value, constant expression, selector which doesn't rely on attributes or local content.

pdpserver/server/ctrl_storage.go
```diff
@@ -101,7 +101,9 @@ func (handler *storageHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 
 	switch cmd {
 	case queryCmd:
+		handler.s.RLock()
```
vasili-v: That's too much to lock. With the lock you can't update policy during whole request handling. You need to lock only storage pointer read (just one quick operation):
```go
handler.s.RLock()
p := handler.s.p
handler.s.RUnlock()
handleQuery(w, p, resourcePath, urlQuery)
```
Read access for storage itself doesn't require locking as it protected by copy-on-write mechanism. What is important to protect is read the pointer to storage.

mingkaic:
@yuewko @rdrozhdzh can you take a look at these changes?

mingkaic:
Since vasili was ok with the changes before he left. I'll merge since there are no further comments

PR #222, opened 2018-11-12
String list and set functions

markinos:
Expands Themis types _List of Strings_ and _Set of Strings_ with more functions:
- ListOfStrings.Contains(String) Boolean
- ListOfStrings.Equal(ListOfStrings) Boolean _(order-dependent)_
- ListOfStrings.Intersect(ListOfStrings) ListOfStrings _(order-agnostic)_
- ListOfStrings.Len() Integer
- SetOfStrings.Equal(SetOfStrings) Boolean
- SetOfStrings.Intersect(SetOfStrings) SetOfStrings
- SetOfStrings.Len() Integer

Unit tests for all new functions (and some previously uncovered older ones) included


pdp/expr_list_of_strings_intersect.go
```diff
@@ -0,0 +1,85 @@
...
+func (f functionListOfStringsIntersect) Calculate(ctx *Context) (AttributeValue, error) {
+	first, err := ctx.calculateListOfStringsExpression(f.first)
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "first argument"), f.describe())
+	}
+	second, err := ctx.calculateListOfStringsExpression(f.second)
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "second argument"), f.describe())
+	}
+	sort.Strings(first)
```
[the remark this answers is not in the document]
yuewko: >  Or you can create map from one of the arguments and use another one as is.

I like this approach.

pdp/expr_stringset_test.go
```diff
@@ -0,0 +1,304 @@
...
+		},
+		{
+			a: newStrTree("foo", "bar", "boo", "boo", "mar", "foo", "boo", "foo"),
+			b: "mar",
+			c: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("Set of Strings Intersect %v + %v", tc.a, tc.b), func(t *testing.T) {
```
yuewko:
```suggestion
		t.Run(fmt.Sprintf("Set of Strings Contains %v + %v", tc.a, tc.b), func(t *testing.T) {
```

pdp/expr_stringset_test.go
```diff
@@ -0,0 +1,304 @@
...
+		},
+		{
+			a: newStrTree("1", "2", "3", "5", "4"),
+			b: newStrTree("4", "3", "1", "2", "5"),
+			c: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("Set of Strings Intersect %v + %v", tc.a, tc.b), func(t *testing.T) {
```
yuewko:
```suggestion
		t.Run(fmt.Sprintf("Set of Strings Equal %v + %v", tc.a, tc.b), func(t *testing.T) {
```

pdp/expr_list_of_strings_test.go
```diff
@@ -212,6 +215,269 @@ func TestFunctionListOfStrings(t *testing.T) {
...
+		},
+		{
+			a: []string{"foo", "foo", "bar", "bar", "bar"},
+			b: []string{"bar", "foo", "foo", "moo"},
+			c: []string{"bar", "foo", "foo"},
+		},
+		{
+			a: []string{"1", "2", "2", "3", "3", "3", "4", "4", "4", "4"},
+			b: []string{"4", "3", "3", "2", "2", "2", "1", "1", "1", "1"},
+			c: []string{"4", "3", "2", "1", "2", "3"},
```
yuewko: I did not envision this is how the list of strings intersect function would behave.

- Does this behavior satisfy your use case? Maybe you can provide a couple of examples?
- I was envisioning that it behaves similarly to set of strings, except that the result is not sorted.

pdp/expr_list_of_strings_intersect.go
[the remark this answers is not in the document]
markinos: I will switch to using a map from one of the arguments and using the other as is to determine intersect values for list of strings. This should simplify the algorithm greatly and will take care of most of Vasili's comments.

pdp/expr_list_of_strings_test.go
markinos: I have reviewed the use cases and I don't think number of repetitions is necessary to store. Changing the intersect algorithm to what you and Vasili recommended above will resolve this behavior.

pdp/expr_list_of_strings_intersect.go
```diff
@@ -0,0 +1,85 @@
...
+	sort.Strings(second)
+	iLen, jLen := len(first), len(second)
+	var res = make([]string, int(math.Max(float64(iLen), float64(jLen))))
```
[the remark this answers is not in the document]
markinos: Changed to Min

pdp/expr_list_of_strings_intersect.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "second argument"), f.describe())
+	}
+
+	values := make(map[string]bool)
+	for _, f := range first {
+		values[f] = false
+	}
+
+	res := make([]string, int(math.Min(float64(len(first)), float64(len(second)))))
```
yuewko: Suggest you do:

`res := make([]string, 0, int(math.Min(float64(len(first)), float64(len(second)))))`

then, you can get rid of `k`

and you can just use append on line 56, i.e. `res = append(res, s)`

pdp/expr_list_of_strings_intersect.go
```diff
@@ -0,0 +1,69 @@
...
+	first, err := ctx.calculateListOfStringsExpression(f.first)
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "first argument"), f.describe())
+	}
+	second, err := ctx.calculateListOfStringsExpression(f.second)
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "second argument"), f.describe())
+	}
+
+	values := make(map[string]bool)
```
yuewko: Since you are not actually using the value `bool`, you can do the following instead:

`values := make(map[string]struct{})`

and the assignment a couple of lines below can just be `values[f] = struct{}{}`

markinos:
@yuewko , I have updated the code based on your inline suggestion as well as included some brief readme updates. Do you think it is worth the time to make more extensive explanation for the new functions (such as adding this functionality to examples/ or by detailing policies in the readme)?

yuewko:
@markinos , your update looks good!

As for documentation, I believe what you added to README is good enough for now.

LGTM, overall!!

pdp/expr_list_of_strings_intersect.go
```diff
@@ -0,0 +1,67 @@
...
```
vasili-v: You can make the map with `make(map[string]bool, len(first))` which as I can recall allows to take less memory by single allocation. Especially in case of small lists.

pdp/expr_list_of_strings_intersect.go
```diff
@@ -0,0 +1,67 @@
...
+	for _, f := range first {
+		values[f] = false
+	}
+
+	res := make([]string, 0, int(math.Min(float64(len(first)), float64(len(second)))))
```
vasili-v: Here using `len(values)` instead of `len(first)` may save few bytes as well.

yuewko:
@vasili-v 
> But I'm still curious if the intersection for lists is absolutely necessary.

I just wanted to point out that the initial reason I suggested implementing intersect for list of strings is because I was assuming that it will be more efficient - I understood that the use case is for relatively small lists and no ordering is required.

If you believe that my assumption is wrong, then I'm all for considering reusing the set of strings implementation.

vasili-v:
@yuewko

> reusing the set of strings implementation

Yes, instead of "intersect" for a couple of list of strings I'd recommend:
- **set of strings** function for any number of "list/set of strings" arguments which produces "set of strings" from all items of given arguments (similar to "list of strings" function);
- **intersect** - for any following combination of arguments:
  - **set of strings** and **set of strings**;
  - **set of strings** and **list of strings**;
  - **list of strings** and **set of strings**.

The intersect in any case returns **set of strings**. It can be implemented as single expression (and piece of code) which works for all the cases.

pdp/expr_stringset_intersect.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "first argument"), f.describe())
+	}
+	secondSet, err := ctx.calculateSetOfStringsExpression(f.second)
+	if err != nil {
+		return UndefinedValue, bindError(bindError(err, "second argument"), f.describe())
+	}
+
+	res := strtree.NewTree()
+	first := firstSet.Enumerate()
```
vasili-v: Sorry, for late comment to the code. You don't need to enumerate both trees. You need to enumerate one and check other:
```go
res := strtree.NewTree()
for p := range firstSet.Enumerate() {
	if _, ok := secondSet.Get(p.Key); ok {
		res.InplaceInsert(p.Key, p.Value)
	}
}
```

It's a bit slower like N*log(M) instead of max(N, M) but I don't think that it's critical to the moment.

vasili-v: Here, you can process cases when one of arguments is list of strings similarly.

vasili-v: One more thing about set of strings. Themis expects value for strtree key-value pair to be integer and tries to order strings in a set by the value where it needs ordered set of strings.

yuewko:
@vasili-v 

> The intersect in any case returns set of strings. It can be implemented as single expression (and piece of code) which works for all the cases.

Sounds like we are in agreement here.
However, I propose that all the enhancements you mentioned in that comment be deferred. @markinos  can create an issue for that so it won't be forgotten.

I think it's safe to defer them because the enhancements will be backward compatible with what the current implementation. And, I believe the current implementation is "good enough" - after addressing your other comments https://github.com/infobloxopen/themis/pull/222#pullrequestreview-175054963

Would that be acceptable to you?

yuewko:
@markinos , please don't forget to create issue for deferred enhancements proposed by @vasili-v . Thanks!

yuewko:
@markinos , thanks for creating the issue (https://github.com/infobloxopen/themis/issues/223).

markinos:
Welcome, @yuewko ! @vasili-v  along with the issue filed, the new equals and contains functions were added to target.go

pdp/target.go
```diff
@@ -196,7 +196,11 @@ var TargetCompatibleExpressions = map[string]map[Type]map[Type]twoArgumentsFunct
 		TypeInteger: {
 			TypeInteger: makeFunctionIntegerEqual},
 		TypeFloat: {
-			TypeFloat: makeFunctionFloatEqual}},
+			TypeFloat: makeFunctionFloatEqual},
```
vasili-v: This won't work. You registered function ListOfStringsEqual which takes two lists of strings as arguments as a function which suitable for TypeListOfStrings and TypeInteger argument pair. Similarly, makeFunctionSetOfStringsEqual won't work with TypeSetOfStrings and TypeFloat pair. Please remove all the new "equal" functions from here or put correct argument types:
```go
var TargetCompatibleExpressions = map[string]map[Type]map[Type]twoArgumentsFunctionType{
	"equal": {
		...
		TypeListOfStrings: {
 			TypeListOfStrings: makeFunctionListOfStringsEqual},
		TypeSetOfStrings: {
			TypeSetOfStrings: makeFunctionSetOfStringsEqual}},
	},
	...
}
```

pdp/target.go
```diff
@@ -209,6 +213,8 @@ var TargetCompatibleExpressions = map[string]map[Type]map[Type]twoArgumentsFunct
 			TypeNetwork: makeFunctionNetworkAddressContainedByNetwork},
 		TypeNetwork: {
 			TypeAddress: makeFunctionNetworkContainsAddress},
+		TypeListOfStrings: {
```
vasili-v: This is ok.

pdp/target.go
vasili-v: Also you can make following change in `github.com/infobloxopen/themis/pdp/ast/yast/parser_test.go` (parset_test.go contains special policy `allFeaturePolicies` to test all possible cases):
```
$ git diff
diff --git a/pdp/ast/yast/parser_test.go b/pdp/ast/yast/parser_test.go
index 077bee5d..3f30ec46 100644
--- a/pdp/ast/yast/parser_test.go
+++ b/pdp/ast/yast/parser_test.go
@@ -204,6 +204,16 @@ policies:
       - val:
          type: string
          content: string
+    - equal:
+      - attr: lsAttr
+      - val:
+         type: list of strings
+         content: [first, second, third]
+    - equal:
+      - attr: ssAttr
+      - val:
+         type: set of strings
+         content: [first, second, third]
     - all:
       - contains:
         - val:
```
With the change I get following in the directory:
```
$ go test -run TestUnmarshal
--- FAIL: TestUnmarshal (0.00s)
    parser_test.go:935: Expected no error but got *yast.matchFunctionCastError (#24 (hidden policy set>target>2>any>3): No function equal for arguments List of Strings and List of Strings)
FAIL
exit status 1
FAIL	github.com/infobloxopen/themis/pdp/ast/yast	0.037s
```
This is exactly because I don't have "equal" function to those arguments. With your changes it should work.

Similar change you can make in `github.com/infobloxopen/themis/pdp/ast/jast/parser_test.go` and check if JAST parser understands you correctly:
```
$ git diff
diff --git a/pdp/ast/jast/parser_test.go b/pdp/ast/jast/parser_test.go
index cca078a0..1fed525f 100644
--- a/pdp/ast/jast/parser_test.go
+++ b/pdp/ast/jast/parser_test.go
@@ -323,6 +323,32 @@ const (
               }
             ]
           },
+          {
+            "equal": [
+              {
+                "attr": "lsAttr"
+              },
+              {
+                "val": {
+                  "type": "list of strings",
+                  "content": ["first", "second", "third"]
+                }
+              }
+            ]
+          },
+          {
+            "equal": [
+              {
+                "attr": "ssAttr"
+              },
+              {
+                "val": {
+                  "type": "set of strings",
+                  "content": ["first", "second", "third"]
+                }
+              }
+            ]
+          },
           {
             "all": [
               {
```

markinos:
@vasili-v thank you for the details on how to properly setup target expressions and the parser tests. I've updated the entries and added the tests you suggested and everything makes fine now.

vasili-v:
Thanks! LGTM!

PR #224, opened 2018-11-30
Pepcli update: list of strings and cl-args

markinos:
Updates pepcli to handle list of strings and add some command line functionality:
- 	pepcli now can marshal list of string type and guess for list of strings
- 	pepcli takes in JSON or YAML filepaths or raw JSON for “-i” arg (fully backwards compatible)
- 	pepcli test package has flagparser option to sort list of strings returned in pdp response obligations
- 	examples package now has 12-lists, which covers a few brief pepcli use cases

New functionality can be tested with these generic commands (json not provided in PR):
- `pepcli -i '{"attributes”:{…},”requests”:[{…},…]}' test/perf` [raw json]
- `pepcli -i requests.json test/perf` [json filepath]
- `pepcli -i requests.yaml test -sort` [sort option]

### DIFF
--- a/pepcli/requests/requests.go
+++ b/pepcli/requests/requests.go
@@ -22,7 +22,7 @@ type requests struct {
 	Requests   []map[string]interface{}
 }
 
-// Load reads given YAML file and porduces list of requests to run.
+// Load reads given YAML file and produces list of requests to run.
 func Load(name string, size uint32) ([]pb.Msg, error) {
 	b, err := ioutil.ReadFile(name)
 	if err != nil {
@@ -80,7 +80,8 @@ var marshallers = map[pdp.Type]attributeMarshaller{
 	pdp.TypeFloat:   floatMarshaller,
 	pdp.TypeAddress: addressMarshaller,
 	pdp.TypeNetwork: networkMarshaller,
-	pdp.TypeDomain:  domainMarshaller}
+	pdp.TypeDomain:  domainMarshaller,
+	pdp.TypeListOfStrings: listOfStringsMarshaller}
 
 func makeAttribute(name string, value interface{}, symbols map[string]pdp.Type) (pdp.AttributeAssignment, error) {
 	t, ok := symbols[name]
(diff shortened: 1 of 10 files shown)


pepcli/requests/requests.go
```diff
@@ -17,22 +19,49 @@ import (
...
-
+// Load reads given data--if it is a filepath that ends in a yaml or json extension and can be read,
+// the respective unmarshaler will be used; otherwise, the input is processed as raw JSON.
+func Load(data string, size uint32) ([]pb.Msg, error) {
 	in := &requests{}
-	err = yaml.Unmarshal(b, in)
-	if err != nil {
-		return nil, err
+
+	switch strings.TrimLeft(strings.ToLower(filepath.Ext(data)), ".") {
```
vasili-v: Actually, JSON formatted string can represent a path in file system. I don't expect anyone to abuse the utility with such corner cases. But it's better to mention somewhere in documentation that strings with .json or .yaml suffix are treated as file names only while anything else will be parsed as JSON.

pepcli/requests/requests.go
```diff
@@ -265,3 +303,25 @@ func domainMarshaller(value interface{}) (pdp.AttributeValue, error) {
...
+	v, ok := value.([]interface{})
+	if !ok {
+		return pdp.UndefinedValue, fmt.Errorf("can't marshal %T as list of string", value)
+	}
+	if len(v) == 0 {
+		return pdp.MakeListOfStringsValue([]string{}), nil
+	}
+
+	los := make([]string, len(v))
+	for _, s := range v {
```
vasili-v: It's better to mentioned wrong item position for example in following way:
```go
for i, s := range v {
	switch value := s.(type) {
	case string:
		los = append(los, value)
	default:
		return pdp.UndefinedValue, fmt.Errorf("can't marshal %T at %d as string in list of strings", s, i+1)
	}
}
```

pepcli/test/test.go
```diff
@@ -95,3 +102,12 @@ func dump(r pdp.Response, f io.Writer) error {
 	_, err := fmt.Fprintf(f, "%s\n", strings.Join(lines, "\n"))
 	return err
 }
+
+func sortDelimitedListOfStrings(unsortedCommaList string, delimiter string) string {
+	list := strings.Split(unsortedCommaList, delimiter)
```
vasili-v: The list is JSON formatted string list with no brackets around (as I can recall). To split it correctly you can add pair of "[" "]" and unmarshal it with json package to []string.

pepcli/requests/requests.go
markinos: That is good to note about JSON formatted string representing a filepath. I've now referenced explicitly how input is parsed in `README.md` and `pepcli/README.md`

markinos:
@vasili-v I've amended PR based your comments. @yuewko has indicted the PR is fine as well, so it should be good to merge

PR #228, opened 2019-01-31
pepcli string list marshaller fix

markinos:
pepcli’s list of strings function always included the first few fields as empty “” (this is because the list was appended after being created); this led to false positives in anything that checked intersection, len, or contents of a list of strings. This update changes the list of strings marshaller to properly reflect the actual values intended without extra empty values.


yuewko:
LGTM

PR #232, opened 2019-04-09
Implement OwnEndpoint such that query returns own IP

onkarbanerjee:


contrib/coredns/policy/policy.go
```diff
@@ -92,6 +93,45 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
 		p.attrPool.Put(attrBuffer)
 	}()
 
+	if p.conf.ownIPEndpoint != "" {
+		var buff bytes.Buffer
+		buff.Write([]byte(p.conf.ownIPEndpoint))
+		buff.Write([]byte("."))
+		buff.Write([]byte(p.conf.debugSuffix))
```
onkarbanerjee: Any better way to do this, maybe?

rdrozhdzh: concatenate it once (in parsing code) then store in config

contrib/coredns/policy/policy.go
```diff
@@ -92,6 +93,45 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
...
+
+		if r != nil && len(r.Question) > 0 && r.Question[0].Name == buff.String() {
+			var rr dns.RR
+			srcIP := getRemoteIP(w)
```
rdrozhdzh: create `func ip2rr(ip net.IP) dns.RR` which will create A or AAAA RR from IP address. This func should be reused in `setRedirectQueryAnswer()`

contrib/coredns/policy/policy.go
```diff
@@ -92,6 +93,45 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
 		p.attrPool.Put(attrBuffer)
 	}()
 
+	if p.conf.ownIPEndpoint != "" {
+		var buff bytes.Buffer
+		buff.Write([]byte(p.conf.ownIPEndpoint))
+		buff.Write([]byte("."))
+		buff.Write([]byte(p.conf.debugSuffix))
+
+		if r != nil && len(r.Question) > 0 && r.Question[0].Name == buff.String() {
```
rdrozhdzh: move this code to a separate func, keep ServeDNS as short as possible

rdrozhdzh: do you care about query type and query class?

PR #233, opened 2019-04-09
Implement OwnEndpoint such that query returns own IP

onkarbanerjee:


contrib/coredns/policy/config.go
```diff
@@ -53,6 +55,13 @@ func policyParse(c *caddy.Controller) (*policyPlugin, error) {
 					return nil, err
 				}
 			}
+			if len(p.conf.ownIPEndpoint) > 0 && len(p.conf.debugSuffix) > 0 {
+				var buff bytes.Buffer
+				buff.Write([]byte(p.conf.ownIPEndpoint))
+				buff.Write([]byte("."))
+				buff.Write([]byte(p.conf.debugSuffix))
+				p.conf.ownIPEndpoint = buff.String()
```
onkarbanerjee: Should we use a separate param in conf to hold the ownIPEndpoint + debugSuffix? As of now there is no other use of ownIPEndpoint , so i used the same param itself.

rdrozhdzh: yes, that's ok. You can even require from user to provide FQDN, and not depend on debug suffix.
Please also update readme

contrib/coredns/policy/policy.go
```diff
@@ -196,3 +201,29 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
 	status = dns.RcodeServerFailure
 	return dns.RcodeSuccess, errInvalidAction
 }
+
+// checkOwnIPEndpointDebug checks if query made to configured ownIPEndpoint and is debug, then returns the status accordingly
+func (p *policyPlugin) checkOwnIPEndpointDebug(w dns.ResponseWriter, r *dns.Msg) (int, bool) {
+	if strings.HasSuffix(p.conf.ownIPEndpoint, p.conf.debugSuffix) &&
```
rdrozhdzh: don't check this - you guaranteed this in config code

contrib/coredns/policy/policy.go
```diff
@@ -196,3 +201,29 @@ func (p *policyPlugin) ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dn
...
+		r.Question[0].Name == p.conf.ownIPEndpoint {
+
+		srcIP := getRemoteIP(w)
+		if srcIP == nil {
+			return dns.RcodeServerFailure, true
+		}
+
+		qName, qClass := getNameAndClass(r)
+		rr := ip2rr(srcIP)
+		if rr == nil || rr.Header() == nil {
```
rdrozhdzh: don't check rr.Header() != nil

contrib/coredns/policy/policy.go
onkarbanerjee: The feature requires that debug queries to `ownIPEndPoint` should get responded with client's IP.  Corefile could have only `ownIPEndpoint` and no debug suffix. Since we use the same variable to concat the debug prefix to it, we could end up with only ownIPEndpoint and no debug prefix, in which case a non-debug query will also respond with client's IP if we omit this check here.

rdrozhdzh: This check doesn't depend on query itself, so no need to check it each time. You can reset `ownIPEndpoint` to "" at config time in case if some config options are incorrect. The check `r.Question[0].Name == p.conf.ownIPEndpoint` will fail if `ownIPEndpoint == ""`, since query name cannot be empty

PR #235, opened 2019-05-09
ATE probe token

onkarbanerjee:
ATE probe token

### DIFF
--- a/contrib/coredns/policy/dns.go
+++ b/contrib/coredns/policy/dns.go
@@ -4,13 +4,29 @@ import (
 	"context"
 	"errors"
 	"net"
+	"strings"
 
 	"github.com/coredns/coredns/plugin"
 	"github.com/coredns/coredns/plugin/pkg/nonwriter"
 	"github.com/miekg/dns"
 )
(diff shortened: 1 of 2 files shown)


contrib/coredns/policy/dns.go
```diff
@@ -114,42 +140,49 @@ func resetTTL(r *dns.Msg) *dns.Msg {
 }
 
 func (p *policyPlugin) setRedirectQueryAnswer(ctx context.Context, w dns.ResponseWriter, r *dns.Msg, dst string) (int, error) {
+	qName, qType, qClass := getNameTypeAndClass(r)
+	rr, err := getRRByType(dst, qName, qType, qClass)
+	if err != nil {
+		return dns.RcodeServerFailure, err
```
onkarbanerjee: `getRRByType()` doesn't err if RR is to be got by data format. It only errs if RR is to be got by rrCodePrefix format and an error is encountered in the same only

onkarbanerjee:
Branch has infoblox specific detail in name and comments

PR #244, opened 2019-12-03
Add themis-fake

tommyongib:
Purpose:
- Enable developers to work locally on their computers with less remote dependencies.

Description:
- When working on a service that depends on themis, instead of pointing to a real themis endpoint, developers can point to this `themis-fake`, which is deployed in a local k3d on their computers.
- `themis-fake` simply permits all requests.


tommyongib:
In meeting 20191204, decided to close for now until further discussion on sdk, bypassing etc.

PR #245, opened 2020-01-17
take standard gRPC ServerOption and expose gRPC Server

drewwells:
Enables passing grpc interceptors to themis so we can leverage the grpc ecosystem middleware like grpc_prometheus and grpc_logrus.

### DIFF
--- a/pdpserver/server/server.go
+++ b/pdpserver/server/server.go
@@ -35,6 +35,13 @@ type transport struct {
 // Option configures how we set up PDP server.
 type Option func(*options)
 
+// WithServerOption specifies gRPC Server Options
+func WithGRPCOptions(opt ...grpc.ServerOption) Option {
+	return func(o *options) {
+		o.grpcOpts = opt
+	}
+}
+
 // WithLogger returns a Option which sets logger
 func WithLogger(logger *log.Logger) Option {
 	return func(o *options) {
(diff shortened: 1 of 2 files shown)


daniel-garcia:
LGTM

pdpserver/server/server.go
```diff
@@ -249,6 +258,11 @@ func NewServer(opts ...Option) *Server {
 	}
 }
 
+// GRPCServer expose the underlying gRPC Server
+func (s *Server) GRPCServer() *grpc.Server {
+	return s.requests.proto
```
drewwells: There's a race that causes this to be `nil` sometimes, we need to wait on a non-nil server

daniel-garcia: Looks like s.requests.proto is being recreated in LoadPolicies()

https://github.com/infobloxopen/themis/pull/245/files/32161ccd581a27f1984710426a54cf488aef7d92#diff-7e71dc72409d1564ddfa250aac5a9eabR556

Why can't the server be configured at Server creation time? What is changing during LoadPolicies that requires a NewGRPC server to be instantiated?

drewwells: it's not being re-created, it's just being created very late in `Serve()`. I'll see if I can get this moved to the appropriate place `NewServer()`

PR #246, opened 2020-02-07
pep client: break infinite loop in func (c *streamConn) connect()

rdrozhdzh:
In case if streaming client cannot connect to PDP endpoint (for some reason)
the `func (c *streamConn) connect()` gets stuck in infinite loop. The proposed
fix is to break the loop when `func (c *streamConn) closeConn()` is called.

### DIFF
--- a/pep/streaming_connection.go
+++ b/pep/streaming_connection.go
@@ -210,6 +210,14 @@ func (c *streamConn) tryConnect(addr string, tracer opentracing.Tracer, timeout 
 	var conn *grpc.ClientConn
 	conn, err = grpc.DialContext(ctx, addr, opts...)
 	if err != nil {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if c.state == scisClosing {
+			c.state = scisDisconnected
+			err = errStreamConnWrongState
+		}
+
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pep/streaming_connection.go
```diff
@@ -210,6 +210,14 @@ func (c *streamConn) tryConnect(addr string, tracer opentracing.Tracer, timeout
...
 	conn, err = grpc.DialContext(ctx, addr, opts...)
 	if err != nil {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+
+		if c.state == scisClosing {
+			c.state = scisDisconnected
+			err = errStreamConnWrongState
+		}
+
```
drewwells: `WithBlock()` would throw an error if the connection to pdp could not be established https://pkg.go.dev/google.golang.org/grpc?tab=doc#DialContext

rdrozhdzh: The intention is to break the loop at https://github.com/infobloxopen/themis/blob/2ae78bb8e45be2c93f2bf9e109e0445395816b9b/pep/streaming_connection.go#L130-L134
This can happen only if `tryConnect()` returns `nil` or `errStreamConnWrongState`

PR #247, opened 2020-03-28
add support for client unary interceptors in pepclient

drewwells:


drewwells:
This passes tests, why is Travis not building the PR?
```
PASS
coverage: 89.7% of statements
ok      github.com/infobloxopen/themis/contrib/coredns/policy   4.338s
```

PR #248, opened 2020-03-28
Feature/client timeouts

drewwells:
This should be backwards compatibility for building, but it will alter some behavior. It is now possible to call Validate() long after the 'Connect()' is called. Connect no longer blocks, and will perform the tcp connection in the background.

It will be possible to cancel ValidateContext() calls where previously this used the confusing 'connTimeout' or 'context' of the initial client setup. was not possible unless the client was built shortly before the Validate() call is made or the confusing 'connTimeout' is used. connTimeout would timeout grpc.Dial and would do things on the validate. 

Validate() continues the same odd behavior of using the connTimeout flag or the client context, neither of these will do what the client is likely to expect so it should not be used.

I didn't touch the Client interface, b/c the streaming client is not generally used and I didn't feel fit to tackle it. Instead users can leverage ValidateContext by type conversion. NewClient().(*UnaryClient).


drewwells:
dependent on #247

PR #251, opened 2020-05-06
return an error, when there is an error

drewwells:


drewwells:
Needs to annotate some common errors with grpc Status codes

pdpserver/server/service.go
```diff
@@ -180,13 +171,14 @@ func (s *Server) Validate(ctx context.Context, in *pb.Msg) (*pb.Msg, error) {
...
 
 	if s.opts.autoResponseSize {
-		msg.Body = s.rawValidate(p, c, in.Body)
+		msg.Body, err = s.rawValidate(p, c, in.Body)
 		return msg, err
 	}
 
 	b := s.pool.Get()
-	msg.Body = s.rawValidateToBuffer(p, c, in.Body, b)
+	body, err := s.rawValidateToBuffer(p, c, in.Body, b)
```
rdrozhdzh, line 179: Here you define new `err` variable, which shadows existing one. Can you make sure that this doesn't break existing logic, e.g. `validatePostHook` logic?

pdpserver/server/stream_service.go
```diff
@@ -55,7 +55,11 @@ func (s *Server) NewValidationStream(stream pb.PDP_NewValidationStreamServer) er
 				return buffer, nil
 			})})
 		} else {
-			err = stream.Send(&pb.Msg{Body: s.rawValidateToBuffer(p, c, in.Body, buffer)})
+			body, err := s.rawValidateToBuffer(p, c, in.Body, buffer)
```
rdrozhdzh, line 58: Here you define new `err` variable, which shadows existing one. Can you make sure that this doesn't break existing logic, e.g. now external `err` is not updated with the error coming from `Send`. Consider renaming the new error variable to avoid ambiguity.

pdpserver/server/service.go
drewwells: fixed

PR #256, opened 2020-10-26
Refreshed vendor tree

UladzimirTrehubenka:
Supersedes incomplete #255 (missing set several endpoints for unary PEP client with resolver/manual object)


contrib/coredns/policy/config.go
```diff
@@ -273,7 +273,7 @@ func (conf *config) parseConnectionTimeout(c *caddy.Controller) error {
 	} else {
 		timeout, err := time.ParseDuration(args[0])
 		if err != nil {
-			return fmt.Errorf("Could not parse timeout: %s", err)
+			return fmt.Errorf("Could not parse timeout: time: invalid duration \"%s\"", args[0])
```
rdrozhdzh, line 276: you can use %q here instead

rdrozhdzh (approved):

PR #271, opened 2023-11-23
Make improvements to themis based on new request format from coredns

d23choi:


pdp/context.go
```diff
@@ -105,7 +105,7 @@ func NewContext(c *LocalContentStorage, count int, f func(i int) (string, Attrib
 // Requirements for the storage are the same as NewContext function has.
 // The request umarshaled to sequence of attributes as descirbed by
 // (Un)MarshalRequest* functions help.
-func NewContextFromBytes(c *LocalContentStorage, b []byte) (*Context, error) {
+func NewContextFromBytes(c *LocalContentStorage, b []byte, currentContext *Context) (*Context, error) {
```
daniel-garcia, line 108: This changes the function signature which could break other clients of this code. Can you make a new method instead?

pdp/context.go
```diff
@@ -117,21 +117,76 @@ func NewContextFromBytes(c *LocalContentStorage, b []byte) (*Context, error) {
...
 
-	ctx := &Context{a: make(map[string]interface{}, count), c: c}
-	for i := 0; i < count; i++ {
-		ID, v, n, err := getRequestAttribute(b[off:])
-		if err != nil {
-			return nil, bindErrorf(err, "%d", i+1)
+	var previousContext *Context
+
+	// Create a new context when we call function for the first time in for loop
+	if currentContext == nil {
```
daniel-garcia, line 123: This new behavior needs a test.

daniel-garcia (changes requested):

